PR #5, opened 2020-05-30
Do not run the toolkit container as the root user

jdoss:
And let's support other Container Engines such as [Podman](https://podman.io/)!


scripts/entrypoint.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
```
sfgeorge, line 1: 👼

jcran:
This is awesome, thanks @jdoss!!!

PR #8, opened 2020-07-30
add snyk and update to defender atp

lindambrown:


lib/kdi/kdi_helpers.rb
```diff
@@ -85,14 +85,43 @@ def create_kdi_asset(asset_hash)
...
+	  def create_kdi_asset_vuln(asset_hash, vuln_hash, match_key=nil)
 	    kdi_initialize unless @assets
 
 	    # check to make sure it doesnt exist
+	    if match_key.nil? then
 			a = @assets.select{|a| uniq(a) == uniq(asset_hash) }.first
-			
+		else
+			case match_key
+	      		when "ip_address"
```
jcran, line 98: why split all these out vs doing

a = @assets.select{|a| a[match_key] == asset_hash.fetch(match_key)}.first

tasks/ms_defender_atp/lib/ms_defender_atp_helper.rb
```diff
@@ -17,14 +17,13 @@ def atp_get_machines(token)
...
   end
   
-  def atp_get_vulns(token,machine_id)
-    print "Getting vulns for machine #{machine_id}"
-    atp_query_api = "https://api.securitycenter.microsoft.com/"
+  def atp_get_vulns(token,atp_query_api)
+    print "Getting vulns"
     headers = {'content-type' => 'application/json', 'accept' => 'application/json', 'Authorization' => "Bearer #{token}", 'accept-encoding' => 'identity'}
-    url =  "#{atp_query_api}/api/machines/#{machine_id}/vulnerabilities"
+    url =  "#{atp_query_api}/api/vulnerabilities/machinesVulnerabilities?skip=10000"
```
jcran, line 26: what does the skip parameter do? Should this be passed to the function so we can paginate?

tasks/ms_defender_atp/ms_defender_atp.rb
```diff
@@ -1,40 +1,50 @@
...
     {
-      id: "microsoft_atp",
-      name: "Microsoft ATP",
-      description: "Pulls assets and vulnerabilitiies from Microsoft ATP",
+      id: "ms_defender_atp",
+      name: "MS Defender ATP",
+      description: "Pulls assets and vulnerabilitiies from Microsoft Defenders ATP",
       options: [
-        {:name => "atp_tenant_id", 
-          :type => "strign",
```
jcran, line 16: good catch!

tasks/ms_defender_atp/lib/ms_defender_atp_helper.rb
lindambrown: They don't have pagination, this is what they (microsoft) recommended I use.

lib/kdi/kdi_helpers.rb
lindambrown: duh thanks!

PR #34, opened 2020-10-08
Update mapper.rb

lindambrown:
added support for tagging


lindambrown:
per request from expanse. they wanted additional tag info for the assets.

PR #38, opened 2020-10-08
More tweaks to DF connectors 

jcran:
Getting close to GA, just knocking out customer-reported bugs. Sorry for the lack of specificness/clarity in these PRs, they're just making stuff work for expanse, riskiq and the overarching mapper.


lindambrown (approved):
looks good

PR #42, opened 2020-10-20
changed to batch process assets

lindambrown:
added tag for defenders and changed to batch process assets


dbrothersKenna (commented):
Not familiar with the other changes but the Tag for MSDefenderATP looks good.

dbrothersKenna (approved):

PR #44, opened 2020-10-23
Revert "changed to batch process assets"

dbrothersKenna:
Reverts KennaPublicSamples/toolkit#42

Problems incurred with GWL(CanadaLife) after merge.  Temporarily reverting until next week to sync with Linda.


jgamblin (approved):
Roll Back

PR #49, opened 2020-10-27
Added required annotation and default value to task specific help for tasks  (task=taskname:help)

tmgerhart:
Added required annotation and default value to task specific help for tasks  (task=taskname:help)

This will provide more context for users to create command lines directly from their terminal without referencing the read.md


jgamblin (approved):
Tested this with Toby this morning and it looks good.

PR #51, opened 2020-10-27
clean up of atp batching work

lindambrown:
created a few new methods. Changed things around so that there is a call to ONLY upload files and the method returned the json which holds the file number. Code then gathers all the file ids and at the very end a call is done to run the connector and provides the final array of file ids.


jgamblin (approved):
This looks good to me.

PR #53, opened 2020-11-02
KISS The Containerfile

jgamblin:
This change removes the Kenna user from the container and runs everything as the default user to fix the issues @tmgerhart is running into with the CSV-KDI script.   It also flattens the container file and makes it more readable by removing variables.


tmgerhart (approved):
This looks good.  From discussing with Jerry, the container will be running by the default user starting the container (not root necessarily).

PR #54, opened 2020-11-02
lastest change to atp batching

lindambrown:
allowed for skipping asset check so when loading assets only there isn't a check for dups.


lindambrown:
also, added logic for kdi json files to be deleted after upload as long as debug is set to false.

jgamblin (approved):

PR #74, opened 2020-11-18
Introduce Rubocop @ v.1.3.1

jheuer:
I figured why not make my first toolkit PR a doozy. This adds rubocop linting to the repo, to provide support in adhering to a basic ruby styleguide. Rubocop can be run locally, and will also be run locally via a Github Action powered by super-linter. Note that the current version of super-linter specifies rubocop @ `~> 0.82.0` [1] whereas this repo itself has the most recent. For this reason, 4 cops found in `.rubocop.yml` are currently commented out, with notes about the respective rubocop version required. When sublime-linter is updated to a more current rubocop, we can reactivate those cops accordingly. [2]

The styleguide in `.rubocop.yml` is based roughly off of the one in our conduit repo. I think of this as first baby steps: the "cops" specified are fairly permissive, and there are no guards in place (yet) forcing new code to satisfy rubocop. But I think it's a good start.

I applied some safe rubocop autocorrection options where possible. I also made some basic hand edits when there was no autocorrection available, but stayed away from any large or higher-risk changes.

I used `rubocop --auto-gen-config --exclude-limit 10` to generate the list of TODOs found in `.rubocop_todo.yml`. Because of this, `bundle exec rubocop` will currently report no findings. Over time, we can improve adherence to the styleguide and remove the overrides found in the TODO file.

[1] https://github.com/github/super-linter/blob/7d54559e11c65230d7a96f1b73b7966812f883fa/dependencies/Gemfile#L7
[2] https://github.com/github/super-linter/pull/1013


jcran:
❤️ ❤️ ❤️

lindambrown:
Jeff, so can I add something to sublime to check formatting while I am coding or is this all stuff you find by running a separate task or waiting until you try to merge and the tests fail? I don't want to click 'go' on this until I know what it means and how it effects the folks (like SEs) trying to add stuff to the toolkit.

jheuer:
@lindambrown I personally haven't integrated it into Sublime, but I know there are a few popular approaches to doing that, I will take a look. This looks like one common approach: https://github.com/SublimeLinter/SublimeLinter-rubocop. Not quite as seamless as having it right in the editor, but for quick local feedback I use `bundler exec rubocop`

jgamblin (approved):
These Look Good! Thanks for the PR!

PR #87, opened 2020-12-01
Useless Assignment Rubocop Clean Up

jgamblin:
Rubocop reported over 100 variables that were set that were never used. I comment them all out when possible in this PR.


tasks/security_scorecard/security_scorecard.rb
```diff
@@ -61,7 +61,7 @@ def ssc_issue_to_kdi_asset_hash(issue)
         ###
         if issue["connection_attributes"]
           if issue["connection_attributes"].is_a? Hash
-            port = issue["connection_attributes"]["dst_port"]
+            # port = issue["connection_attributes"]["dst_port"]
```
lindambrown, line 64: is this what it looks like? that the line is commented out?

tasks/snyk_findings/snyk_findings.rb
```diff
@@ -154,14 +143,6 @@ def run(opts)
               "tags" => [project.fetch("source"), project.fetch("packageManager")]
 
             }
```
lindambrown, line 145: these blocks are just removed? I just added them to fix some issues with the locator info

tasks/expanse/lib/client.rb
```diff
@@ -67,9 +67,9 @@ def exposures(max_pages = 100, limit_per_page = 10_000)
         end
 
         def cloud_exposure_counts
-          url = "https://expander.expanse.co/api/v1/summaries/cloud/counts"
-          response_body = RestClient.get(url, @headers)
-          result = JSON.parse(response_body)["data"]
+          # url = "https://expander.expanse.co/api/v1/summaries/cloud/counts"
+          # response_body = RestClient.get(url, @headers)
+          # result = JSON.parse(response_body)["data"]
```
lindambrown, line 72: do we not need this method? looks like the entire thing is being commented out?

tasks/security_scorecard/security_scorecard.rb
jgamblin: Yes,  it was set but not called anywhere else.  I just didn't delete the lines in case they were needed later.

tasks/snyk_findings/snyk_findings.rb
jgamblin: Where in the code did you call tags again?  Rubocop says they are not used.

tasks/expanse/lib/client.rb
jgamblin: At some point, we should delete all the unused lines but I just commented them out.

tasks/bitsight/spec/bitsight_spec.rb
```diff
@@ -141,11 +141,11 @@
       end
 
       it "should translate api results into vali kdi" do
-        guid = get_my_company(@key)
+        # guid = get_my_company(@key)
 
-        results = get_bitsight_findings_and_create_kdi(@key, guid, 100)
+        # results = get_bitsight_findings_and_create_kdi(@key, guid, 100)
```
lindambrown, line 146: I think these 2 lines are needed but the second one does not need to return back to a variable.

tasks/riskiq/riskiq.rb
```diff
@@ -96,26 +96,26 @@ def run(options)
...
+          # result = client.search_global_inventory(client.cve_footprint_query, max_pages)
+          # output = convert_riq_output_to_kdi result
         end
 
         if @riq_create_open_ports # 156220
           print_good "Getting open ports from footprint... BETA feature!"
-          result = client.search_global_inventory(client.open_port_query, max_pages)
-          output = convert_riq_output_to_kdi result
+          # result = client.search_global_inventory(client.open_port_query, max_pages)
+          # output = convert_riq_output_to_kdi result
```
lindambrown, line 112: another one where second line just needs the variable removed?

jheuer:
This is a type of cop where there could definitely be reasonable difference of opinion. I could imagine the argument: I want to assign all of the data points coming in from a 3rd party source to variables, because we're not ready to use all of them yet, but we will one day, and I want it documented how to get at each data point.

That being said, in my view, in almost all cases, an unused variable is more suggestive of a bug, and it's worth checking for.

Keep in mind that we can have it both ways: commit to generally fixing the cop complaints, but in specific cases where it's worth departing from the rule, by wrapping any offending code with:

```
# rubocop:disable Lint/UselessAssignment
...
# rubocop:enable Lint/UselessAssignment
```

tasks/snyk_findings/snyk_findings.rb
lindambrown: tags is added to the asset hash. i'll check the code again.

lindambrown (approved):

PR #91, opened 2020-12-10
Update user_role_sync.rb

dbrothersKenna:
Updated user_role_sync with ability to  exclude a list of user roles within Kenna.


dbrothersKenna:
Lint Errors. Correcting and will resubmit

PR #93, opened 2020-12-15
Docker Clean-up

jgamblin:
These changes include adding a [dockerignore](https://docs.docker.com/engine/reference/builder/#dockerignore-file) file to control what we copy to the container in a better manner. 

Only installing production gems in the container using: `RUN bundle install --without development test`

Not installing `Python3` since there is no python code in the repo.  These changes slims the image down 154MB. 

Screen Shot 2020-12-15 at 1 01 27 PM


lindambrown (approved):
Looks fine

PR #97, opened 2020-12-31
riskiq updates

lindambrown:
streamlined processing, added limiting functions, changed output building for faster process. Added lazy processing for added performance.


lindambrown:
Some low level changes worked in here including a new method to improve performance when writing files. They won't be pretty json but this method eliminates the parse function with is very resource intensive. Should help with any connector that produces large files write_file_stream output_dir, filename, false, @paged_assets, @vuln_defs

jgamblin (approved):

PR #99, opened 2021-01-05
Temporarily Remove Solargraph

jgamblin:
Temporarily remove Solargraph which is used for VSCode dev that has an unfixed CVE.


lindambrown (approved):
looks fine.

PR #104, opened 2021-01-12
A few small optimizations to kdi_helpers

jheuer:
Based on some feedback in #dev_internal [here](https://kennasecurity.slack.com/archives/C012YTRLQNQ/p1610395067112100?thread_ts=1610393051.108900&cid=C012YTRLQNQ) I pulled a few operations out of the `@assets.lazy.find{}` loop, so that they are done only once, not once for each member of `@assets`.

I also consolidated some duplicated code into a small method.

Full disclosure: I did not formally measure the performance improvement here, nor did I retest the functionality.

### DIFF
--- a/lib/kdi/kdi_helpers.rb
+++ b/lib/kdi/kdi_helpers.rb
@@ -55,9 +55,9 @@ module Kenna
 
         # if we already have it, skip ... do this by selecting based on our dedupe field
         # and making sure we don't already have it
-
+        uniq_asset_hash = uniq(asset_hash)
         if dup_check
-          return nil if @assets.lazy.select { |a| uniq(a) == uniq(asset_hash) }.any?
+          return nil if @assets.lazy.select { |a| uniq(a) == uniq_asset_hash }.any?
         end
 
         # create default values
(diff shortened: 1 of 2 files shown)


lib/kdi/kdi_helpers.rb
```diff
@@ -72,6 +72,31 @@ def create_kdi_asset(asset_hash, dup_check = true)
...
+        a = if match_key.nil?
+              @assets.lazy.find { |asset| uniq(asset) == uniq_asset_hash }
+            else
+              @assets.lazy.find { |asset| asset[match_key] == asset_hash_key }
+            end
+
+        # SAnity check to make sure we are pushing data into the correct asset
+        unless a # && asset[:vulns].select{|v| v[:scanner_identifier] == args[:scanner_identifier] }.empty?
+          puts "Unable to find asset #{asset_hash}, creating a new one... "
+          create_kdi_asset(asset_hash)
```
bschmeck, line 89: You can pass `true` here as the second argument to `create_kdi_asset` to skip the lookup in that method, since we already tried to look up the asset and know that it's missing.

lindambrown: you meant pass false right? dup_check = false because we already know the asset exists? I added this boolean.

lindambrown (approved):

lib/kdi/kdi_helpers.rb
bschmeck: :facepalm: True, I did mean `false` :stuck_out_tongue:

PR #110, opened 2021-01-22
new tvm api

lindambrown:


tmgerhart (approved):

tasks/ms_defender_tvm/lib/ms_defender_tvm_helper.rb
```diff
@@ -0,0 +1,140 @@
...
+      def connector_kickoff(kenna_connector_id, kenna_api_host, kenna_api_key, max_retries = 3)
+        ### Finish by uploading if we're all configured
+        return unless kenna_connector_id && kenna_api_host && kenna_api_key
+
+        print_good "Attempting to run to Kenna Connector at #{kenna_api_host}"
+        run_files_on_kenna_connector kenna_connector_id, kenna_api_host, kenna_api_key, @uploaded_files, max_retries
+      end
+
+      def tvm_get_machines(page_param = nil)
+        print_debug "Getting machines"
```
rick-kenna, line 42: Do you want this and other print_debug() statements left in the code?

lindambrown: yeah, they will only print out if debug=true so in most cases users will never see them.

PR #135, opened 2021-03-19
UserAgent

jgamblin:
Add Toolkit UserAgent


jgamblin:
I changed the toolkit user agent to "Kenna Toolkit" so tracking it should be spot on.  If someone with more ruby skills than me could figure out how to pass the task variable to, that would be the end goal of this PR.  For now, it will help with data collection and tracking.

lindambrown (approved):

PR #151, opened 2021-04-13
Added Ext ID to Asset

dbrothersKenna:
Added Ext ID to Asset


katieweb (approved):

tasks/connectors/veracode_asset_vulns/lib/veracode_av_client.rb
```diff
@@ -87,12 +87,15 @@ def issues(app_guid, app_name, tags, page_size)
               # IF "STATIC" SCAN USE FILE, IF "DYNAMIC" USE URL
               file = nil
               url = nil
+              ext_id = nil
```
rick-kenna, line 90: I see this is set to nil, but not checked anywhere in your changes, so do you expect it to just show nil somewhere?

PR #164, opened 2021-05-04
Fixes to CVE Identifiers

dbrothersKenna:
Fixes to CVE Identifiers


lindambrown (commented):
in the vuln def section do you want to be more explicit about when cwe is added? you only want it added if there is no CVE right? Only one things gets saved and we found with some other code that adding everything you have (cve, cwe, wasc) gives undesirable results.

lindambrown (approved):

PR #170, opened 2021-05-07
Add Edgescan connector task

edblaz:
Added some code for the Edgescan connector task. Feel free to reach out to me here or via email (edvardas.b@edgescan.com) any time if there's anything I could help with.


tasks/connectors/edgescan/lib/edgescan_definition.rb
```diff
@@ -0,0 +1,37 @@
...
+            "description" => data["description_src"],
+            "solution" => data["remediation_src"],
+            "cve_identifiers" => cves,
+            "cwe_identifiers" => cwes
+          }.compact
+        end
+
+        private
+
+        def cves
```
lindambrown, line 27: It's better to set a precedence for the identifiers and send in only the one you want used. So if cve send cve, only send cwe if no cve. Make sense?

tasks/connectors/edgescan/lib/kenna_api.rb
```diff
@@ -0,0 +1,112 @@
...
+          kdi_upload(@output_dir, "batch-#{millis}.json", @kenna_connector_id, @kenna_api_host, @kenna_api_key)
+        end
+
+        # Kicks off connector tasks so that whatever was uploaded actually gets imported into Kenna
+        def kickoff
+          kdi_connector_kickoff(@kenna_connector_id, @kenna_api_host, @kenna_api_key)
+        end
+
+        # Fetches existing assets tagged with these Edgescan IDs from the Kenna API
+        def fetch_assets_with_edgescan_ids(edgescan_ids)
```
lindambrown, line 57: let's talk about why this is needed. most of the toolkit connectors don't pull assets from Kenna. Kenna will automatically match up assets in an inbound KDI file to existing assets in Kenna.

tasks/connectors/edgescan/lib/edgescan_definition.rb
edblaz: Makes perfect sense. I made that change in this commit: https://github.com/KennaSecurity/toolkit/pull/170/commits/4da1aab47d8a0dd8a7b7d62516556fde143b27b7

Closing this for now but feel free to re-open if you think I should have gone another route

edblaz:
Hey, 

I just pushed a few more changes:
- The task is no longer explicitly marking assets for removal and instead letting Kenna handle that automatically 
- Added precedence when sending CVE and CWE identifiers
- After learning about automatic inactive asset deletion, removed the ability to sync _only updates_ from a certain timestamp on
- Added the ability to provide what Edgescan API hostname you want used in case you guys ever need to hit `demo.edgescan.com` 

Let me know if you have any gripes or questions!

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
```diff
@@ -0,0 +1,64 @@
...
+
+        def to_kenna_vulnerability
+          {
+            "scanner_type" => "Edgescan",
+            "scanner_identifier" => data["definition_id"],
+            "created_at" => data["created_at"],
+            "last_seen_at" => data["updated_at"],
+            "scanner_score" => data["threat"],
+            "score_map" => { "5" => "10", "4" => "8", "3" => "6", "2" => "4", "1" => "2" },
+            "status" => data["status"],
```
lindambrown, line 34: sorry I missed this the first time but it looks like all the scanner scores I can see are under 5. score_map isn't a field in the KDI so it is likely just getting ignored. Can you use the map to update the value you have in scanner_score?

lindambrown: @edblaz

edblaz: It's great that you caught that! 

The `threat` on Edgescan will always be an integer between 1 and 5. That means that what we're supplying to Kenna will always be between 2 and 10. I hope that works!

lindambrown (approved):

PR #179, opened 2021-05-26
Adding Contrast connector

davidaustinarcher:
This PR includes a change to toolkit.rb as the equals symbol can be included within the contrast_auth_token parameter which can be in the format "contrast_auth_token=ZGF2aWQ=" for example. The change should not affect other users and the value for the parameter should include all the text after the first equals symbol.


tasks/connectors/contrast/contrast.rb
```diff
@@ -0,0 +1,425 @@
...
+              description: "Filter vulnerabilities and libraries using a comma separated list of application tags" },
+            { name: "contrast_include_vulns",
+              type: "boolean",
+              required: false,
+              default: true,
+              description: "Include vulnerabilities from Contrast Assess" },
+            { name: "contrast_environments",
+              type: "string",
+              required: false,
+              default: "",
```
lindambrown, line 58: for environments and severities, will nil value return all results?

lindambrown: @davidaustinarcher

tasks/connectors/contrast/contrast.rb
```diff
@@ -0,0 +1,425 @@
...
+          end
+        end
+
+        if results == true
+          ### Write KDI format
+          kdi_output = { skip_autoclose: false, assets: @assets, vuln_defs: @vuln_defs }
+          output_dir = "#{$basedir}/#{@options[:output_directory]}"
+          filename = "generator.kdi.json"
+          write_file output_dir, filename, JSON.pretty_generate(kdi_output)
+          print_good "Output is available at: #{output_dir}/#{filename}"
```
lindambrown, line 296: can we change this to: 
kdi_upload @output_dir, filename, @kenna_connector_id, @kenna_api_host, @kenna_api_key, false, 3, 1

it will write the file out with a newer more efficient streaming method and automatically upload if the kenna params are not nil
@davidaustinarcher

lindambrown, line 296: please switch to write_file_stream(directory, filename, autoclose, assets, vuln_defs, version = 1)

to replace lines 292-295 we have found the json parsing to be very inefficient

tasks/connectors/contrast/readme.md
```diff
@@ -0,0 +1,59 @@
...
+
+```bash 
+    docker run -it --rm toolkit:latest \
+    task=contrast \
+    contrast_host=<your host> \
+    contrast_org_id=<your org> \
+    contrast_api_key=<your api key> \
+    contrast_auth_token=<your auth header> \
+    kenna_api_key=<your kenna api key> \
+    kenna_api_host=<your kenna api environment, e.g. xxx.kennasecurity.com> \
```
lindambrown, line 37: host isn't needed. it's always api.kennasecurity.com unless the customer is on a dedicated private cloud.

toolkit.rb
```diff
@@ -11,7 +11,7 @@
 args = {}
 args_array.each do |arg|
   arg_name  = arg.split("=").first.to_sym
-  arg_value = arg.split("=").last
+  arg_value = arg.gsub(arg.split("=").first, "").delete_prefix("=")
```
lindambrown, line 14: was there a special case that this fixed? We haven't had any report problems with the current arg logic...

tasks/connectors/contrast/contrast.rb
```diff
@@ -0,0 +1,425 @@
+require_relative "lib/client"
+require "time"
```
lindambrown, line 2: i thought time was included with ruby, is the require explicitly needed?

tasks/connectors/contrast/contrast.rb
davidaustinarcher: Yes that is correct

tasks/connectors/contrast/contrast.rb
davidaustinarcher: Confirmed, I removed this.

tasks/connectors/contrast/readme.md
davidaustinarcher: Removed.

tasks/connectors/contrast/contrast.rb
davidaustinarcher: Resolved.

toolkit.rb
sfgeorge, line 14: Thank you for finding and fixing this issue with multiple equal signs!

Unfortunately, this is a dicey one to fix, and I'm afraid that even the solution has its own unique problems.  For instance, if the unparsed `arg` in this case were the string `foo=foo`, then this resulting substitution would produce an empty string.

I think what we should ultimately do here is optionally support quoted-strings, so that argument values including quotes and spaces can be honored appropriately.

You've done a good job trying to improve our homemade command-line parsing code here (thank you!), but in order to do this parsing safely, I believe that switching to a supported library may be a good approach here.  For instance, if we move to the Ruby standard library [OptionParser](https://ruby-doc.org/stdlib-2.6.6/libdoc/optparse/rdoc/OptionParser.html), it handles a lot of the complexity for us here for free.

davidaustinarcher: No worries, let me know what you decide to go with and we'll fit in - thanks!

lindambrown: I think quoting the inbound params works. That is what we have done with other tasks where the keys have = signs in the key.

run_docker.sh
```diff
@@ -0,0 +1,104 @@
+#David Archer Demo
```
lindambrown, line 1: You didn't mean to check this in, did you?

tasks/connectors/contrast/contrast.rb
lindambrown: Can you note that in the read.me file?  @davidaustinarcher

tasks/connectors/contrast/contrast.rb
lindambrown: I haven't tested but will "version = 1" work or should it just be "1". Did you test after this change? @davidaustinarcher

toolkit.rb
lindambrown: @davidaustinarcher can you back this out and retest with quotes around the keys at run? @davidaustinarcher

davidaustinarcher:
Closing this out, will re-open when ready.

PR #187, opened 2021-06-17
Build Lacework into the toolkit

shotop:
This PR adds a connector to the toolkit for retrieving host and vulnerability data from the lacework api.  It pipes that data into Kenna using the KDI and Kenna API.  This is my first foray into the toolkit, so feedback welcome!


cyberhiker (approved):

jcran:
Nice work @shotop!

PR #196, opened 2021-07-06
Modified User-Agent to standard

rick-kenna:
The fix is within the scope.  However, it probably would be useful to get the header in one spot.

Also, I will create a ticket to update the version with GitHub hooks.

### DIFF
--- a/tasks/utilities/user_role_sync/user_role_sync.rb
+++ b/tasks/utilities/user_role_sync/user_role_sync.rb
@@ -83,16 +83,12 @@ module Kenna
         @role_exclusions = @options[:role_exclusions] # || ""
 
         api_client = Kenna::Api::Client.new(@api_token, @api_host)
-        @version = api_client.version
 
         # Variables we'll need later
         @role_post_url = "https://#{@api_host}/roles"
         @user_post_url = "https://#{@api_host}/users"
-        @headers = {
-          "content-type" => "application/json",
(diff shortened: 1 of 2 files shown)


tasks/utilities/user_role_sync/user_role_sync.rb
```diff
@@ -82,10 +82,17 @@ def run(options)
...
+        @version = api_client.version
+
         # Variables we'll need later
         @role_post_url = "https://#{@api_host}/roles"
         @user_post_url = "https://#{@api_host}/users"
-        @headers = { "content-type" => "application/json", "X-Risk-Token" => @api_token, "User-Agent" => "Kenna Toolkit - user_role_sync" }
+        @headers = {
+          "content-type" => "application/json",
+          "X-Risk-Token" => @api_token,
+          "User-Agent" => "Toolkit.user_role_sync/1.0.#{@version} (Kenna Security)"
```
sfgeorge, line 94: I agree it would be good to get the header-building in one spot.  I guess it would need to accept the role-name as a parameter.

Also, It would be good if Kenna::Api::Client knew of the whole MAJOR.MINOR.REVISION of the version string.

rick-kenna: I will put the header in one place, but the version I will change the version when I get the GitHub hook working.

lindambrown (approved):

PR #201, opened 2021-08-20
Add Cobalt.io Connector

ericcornelissen:
This adds a connector/task for Cobalt.io (_full disclosure, I'm a software engineer at Cobalt.io_). It works by pulling in findings and asset information from [the Cobalt.io public API](https://docs.cobalt.io). I based the task implementation and corresponding README off of [Snyk's task](https://github.com/KennaSecurity/toolkit/tree/6733d97af1824e4594518911513e3474711b2173/tasks/connectors/snyk).

Some notes on the task that may require additional attention:

- Findings marked as "wont_fix" and "valid_fix" in Cobalt are imported and flagged as "risk_accepted" and "resolved" respectively. Alternatively these could not be imported at all.
- Findings marked as "duplicate", "out_of_scope", or "invalid" in Cobalt are not imported at all.
- It currently imports all findings for all assets in Cobalt.io. I implemented it this way because that's how Snyk's task seems to work. A reasonable alternative would be to import [findings from one asset or one pentest](https://docs.cobalt.io/#get-specific-findings). Note that in that case there will always be only one Kenna Security asset in the imported dataset.

One open question from my side is whether to import findings as _vulns_ or as _findings_. The difference is not quite clear to me, except that when I import findings from Cobalt as _findings_ I cannot find them in the Kenna Dashboard.

### DIFF
--- a/tasks/connectors/cobaltio/cobaltio.rb
+++ b/tasks/connectors/cobaltio/cobaltio.rb
@@ -1,0 +1,171 @@
+# frozen_string_literal: true
+
+require_relative "lib/cobaltio_helper"
+
+module Kenna
+  module Toolkit
+    class Cobaltio < Kenna::Toolkit::BaseTask
+      include Kenna::Toolkit::CobaltioHelper
+
+      SCANNER = "Cobalt.io"
+
+      def self.metadata
+        {
+          id: "cobaltio",
+          name: "Cobalt.io",
+          description: "Pulls findings from Cobalt.io",
+          options: [
+            { name: "cobalt_api_token",
+              type: "api_key",
+              required: true,
+              default: nil,
+              description: "Cobalt.io API token" },
+            { name: "cobalt_org_token",
+              type: "api_key",
+              required: true,
+              default: nil,
+              description: "Cobalt.io org token" },
+            { name: "retrieve_from",
+              type: "date",
+              required: false,
+              default: 90,
+              description: "default will be 90 days before today" },
+            { name: "projectName_strip_colon",
+              type: "boolean",
+              required: false,
+              default: false,
+              description: "strip colon and following data from Project Name - used as application identifier" },
+            { name: "packageManager_strip_colon",
+              type: "boolean",
+              required: false,
+              default: false,
+              description: "strip colon and following data from packageManager - used in asset file locator" },
+            { name: "package_strip_colon",
+              type: "boolean",
+              required: false,
+              default: false,
+              description: "strip colon and following data from package - used in asset file locator" },
+            { name: "kenna_connector_id",
+              type: "integer",
+              required: false,
+              default: nil,
+              description: "If set, we'll try to upload to this connector" },
+            { name: "kenna_api_key",
+              type: "api_key",
+              required: false,
+              default: nil,
+              description: "Kenna API Key" },
+            { name: "kenna_api_host",
+              type: "hostname",
+              required: false,
+              default: "api.kennasecurity.com",
+              description: "Kenna API Hostname" },
+            { name: "output_directory",
+              type: "filename",
+              required: false,
+              default: "output/cobaltio",
+              description: "If set, will write a file upon completion. Path is relative to #{$basedir}" }
+
+          ]
+        }
+      end
+
+      def run(opts)
+        super # opts -> @options
+
+        cobalt_api_token = @options[:cobalt_api_token]
+        cobalt_org_token = @options[:cobalt_org_token]
+
+        @kenna_api_host = @options[:kenna_api_host]
+        @kenna_api_key = @options[:kenna_api_key]
+        @kenna_connector_id = @options[:kenna_connector_id]
+
+        # output_directory = @options[:output_directory]
+
+        projectName_strip_colon = @options[:projectName_strip_colon]
+        packageManager_strip_colon = @options[:packageManager_strip_colon]
+        package_strip_colon = @options[:package_strip_colon]
+        to_date = Date.today.strftime("%Y-%m-%d")
+        retrieve_from = @options[:retrieve_from]
+        from_date = (Date.today - retrieve_from.to_i).strftime("%Y-%m-%d")
+
+        findings_json = cobalt_get_findings(cobalt_api_token, cobalt_org_token)
+        print_debug "findings json = #{findings_json}"
+
+        severity_map = { "high" => 7, "medium" => 5, "low" => 3 } # converter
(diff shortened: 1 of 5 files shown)


jcran:
Might want to add it into the main README.md as well.

ericcornelissen:
I'm not sure what you're referring to @jcran, I did add it [in the README.md at the root of the project in the "Toolkit Capabilities (TASKS)" section](https://github.com/ericcornelissen/toolkit/tree/task/cobaltio#toolkit-capabilities-tasks)? I didn't see any other list of tasks in that README.md, and wouldn't know what other README.md is the main one.

jcran:
Ah sorry, missed it at first glance. Great work @ericcornelissen!

tasks/connectors/cobaltio/cobaltio.rb
```diff
@@ -0,0 +1,171 @@
...
+            "external_id" => asset_id,
+            "application" => application,
+          }
+
+          vuln_def_name = resource["title"]
+          description = resource["description"]
+          solution = resource.fetch("suggested_fix") if resource.key?("suggested_fix")
+
+          vuln_def = {
+            "scanner_type" => SCANNER,
```
lindambrown, line 118: Scanner type would = Cobalt in your case.

tasks/connectors/cobaltio/cobaltio.rb
```diff
@@ -0,0 +1,171 @@
...
+          resource = finding_obj["resource"]
+          links = finding_obj["links"]
+          log = resource["log"]
+
+          asset_id = resource["asset_id"]
+          cobalt_asset = cobalt_get_asset(cobalt_api_token, cobalt_org_token, asset_id)
+          application = cobalt_asset["resource"]["title"] if cobalt_asset
+
+          asset = {
+            "external_id" => asset_id,
```
lindambrown, line 109: Usually there is a url or file identifier by which a customer would know the object that needs to be fixed. Can you add that as well? or is it just asset_id?

tasks/connectors/cobaltio/cobaltio.rb
```diff
@@ -0,0 +1,171 @@
...
+            "application" => application,
```
lindambrown, line 110: right now the application appears to be the same as the locator see my comments above. Are there different pages(urls) that make up "Saxophone external network"?

tasks/connectors/cobaltio/cobaltio.rb
```diff
@@ -0,0 +1,171 @@
...
+            "last_seen_at" => last_seen_at,
+            "severity" => severity,
+            "vuln_def_name" => vuln_def_name,
+            "triage_state" => triage_state,
+            "additional_fields" => additional_fields,
+          }
+          finding.compact!
+
+          # Create the KDI entries
+          create_kdi_asset_vuln(asset, finding)
```
lindambrown, line 159: this would change to create kdi asset finding if using the findings model

tasks/connectors/cobaltio/cobaltio.rb
```diff
@@ -0,0 +1,171 @@
...
+            { name: "packageManager_strip_colon",
+              type: "boolean",
+              required: false,
+              default: false,
+              description: "strip colon and following data from packageManager - used in asset file locator" },
+            { name: "package_strip_colon",
+              type: "boolean",
+              required: false,
+              default: false,
+              description: "strip colon and following data from package - used in asset file locator" },
```
lindambrown, line 47: doesn't look like you need any of these strip colon params or am I missing something?

tasks/connectors/cobaltio/lib/cobaltio_helper.rb
```diff
@@ -0,0 +1,66 @@
+# frozen_string_literal: true
+
+module Kenna
+  module Toolkit
+    module CobaltioHelper
+      @@assets = []
+
+      def cobalt_get_assets(api_token, org_token)
```
lindambrown, line 8: might need to think about this later. for a large implementation all those assets would be stored in memory for the container engine and can grind processing to a halt

lindambrown:
@ericcornelissen I added a number of comments. We can schedule a call to review if you want. There are essentially 2 models where cwe/app type findings can go, under the VM model used for network assets (servers and such) or under the newer Findings model. It is mostly the same data but the findings model has more the statuses you are looking for. Under the VM model, there are only 2, open and closed. If you want to test with the findings model, you would call the add asset finding method and I would need to turn on that model in your instance as it is currently under development. If you look at Contrast as an example you will see that they made VM vs Findings an option via param. I can create you an additional instance if you want to use findings. A kenna instance can support one or the other for files and URLs.

tasks/connectors/cobaltio/cobaltio.rb
ericcornelissen: At this stage [the API endpoint for assets](https://docs.cobalt.io/#assets) does not provide this kind of data, partially because an asset typically represent more than one url or file identifier. Note that the `links.ui.url` value links to the asset page in the Cobalt web app (if I understand correctly, this is not the URL you're interested in).

Perhaps a better approach would be to use the `affected_targets` value of a finding itself (see [the API endpoint for findings](https://docs.cobalt.io/#findings))? If so, two questions:

1. As this is a list, it could happen that there is no affected target in the API response. What would be your suggestion in this happens? Resort to just the asset_id, omit the finding altogether, or something else?
2. Should both the `url` and the `external_id` be set, or just the `url`?

tasks/connectors/cobaltio/cobaltio.rb
ericcornelissen: Yes, typically an asset will have more than one page, see also my comment above.

tasks/connectors/cobaltio/cobaltio.rb
ericcornelissen: You are right, I forgot to remove this after copying the metadata from Snyk. I will remove this.

tasks/connectors/cobaltio/lib/cobaltio_helper.rb
ericcornelissen: This is a fair concern, though for most users the request to the `/findings` endpoint is a larger concern.

Unfortunately the (fairly new) Cobalt API does not yet support getting a specific asset (same for findings) or a subset of assets. I'll bring this up to our API team.

ericcornelissen:
Thank you for the review @lindambrown. I think a review call would be helpful. Thank you for the explanation, it does explain why I couldn't see the imported findings in our instance. I would indeed agree that Kenna's findings model better suits Cobalt's data. bit since that is still under development it would make sense to provide support for both findings and vulns. From that perspective, having two test instances would be ideal for us.

PR #204, opened 2021-08-24
gen improvements

lindambrown:
if debug is true OR no connector information is provided, files will be saved to the drive. Otherwise, they will be deleted once they are successfully uploaded to Kenna.

also remove a puts in snyk
removed df port mapper which isn't used
added container and image id to asset kdi helper


rick-kenna (approved):
Looks good.

PR #209, opened 2021-09-07
Add support for findings to Cobalt.io task

ericcornelissen:
Following up on #201, this add support for importing data from Cobalt.io as _findings_. This time the implementation is based off of [the Contrast task](https://github.com/KennaSecurity/toolkit/tree/master/tasks/connectors/contrast). Note that I changed the default import from vulns to findings as the latter betters suites the Cobalt data format.

On top of that, I
- removed an unused option (3b0b9c9abc4e8604567e9d94b38ecfd954c5a13d),
- added a missing option to the tasks metadata (ae48bca939a62ca327437f184ed3f2528b31345d), and
- added the `status` and `scanner_score` fields for vulns.

### DIFF
--- a/tasks/connectors/cobaltio/cobaltio.rb
+++ b/tasks/connectors/cobaltio/cobaltio.rb
@@ -25,11 +25,6 @@ module Kenna
               required: true,
               default: nil,
               description: "Cobalt.io org token" },
-            { name: "retrieve_from",
-              type: "date",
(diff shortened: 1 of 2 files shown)


rick-kenna (approved):
The code is OK with me, but you might consider waiting for Linda Brown, David Brothers, or Jerry Gamblin to also review.

PR #226, opened 2021-10-14
Add mapping for veracode owner

dbrothersKenna:


tasks/connectors/veracode_asset_vulns/lib/veracode_av_client.rb
```diff
@@ -369,6 +374,7 @@ def find_missing_kenna_assets(application)
           enc_close_paren = "%29"
           enc_ampersand = "%26"
 
+          # encoding problematic characters for use in call to Kenna API
```
rick-kenna, line 377: Thanks for the comment.

PR #240, opened 2021-10-29
fix CVE field name

lindambrown:


tasks/connectors/wiz/wiz_task.rb
```diff
@@ -136,20 +136,21 @@ def run(opts)
             vuln_score = vuln_severity[severity].to_i
             if abspath.include? "VIRTUAL_MACHINE"
               runtime = row["Runtime"]
+              os = ow["OperatingSystem"] unless row["OperatingSystem"].empty?
```
bschmeck, line 139: Should this be `row` instead of `ow`?

bschmeck (approved):
I think one typo, otherwise :+1:

rick-kenna (approved):

PR #241, opened 2021-11-02
Tweak what information Edgescan sends to Kenna.

edblaz:
# Requirements
image

# Changes in this PR

- Updated `scanner_type` to be either `EdgescanNet` or `EdgescanApp` depending on the layer of the vulnerability.
- Made it so assets no longer set `application`.

# Questions

We also had two quick questions that came up during development.

### Question 1: No more application ID

Just wanted to verify if we read you right for points `a.ii` and `b.ii`:
image

We've changed it so that assets no longer have an application ID. This means that they will no longer be grouped together in the Applications view on the frontend. We've made this change to both the network and application assets. 

### Question 2: Handling IP blocks

Points `a.i` and `b.i` are fine for the majority of cases:
image

We just wanted to double check with you about a rare case. 

Assets on Edgescan can also have IP block locators. For example an asset may have a locator that looks like this: `192.168.0.1-50`. This would mean that the asset has 50 IPs, everything between `192.168.0.1` and `192.168.0.50`.

At the moment, if the connector comes across an Edgescan asset with an IP block it just creates a single Kenna asset with an external ID locator that looks something like this:
```ruby
{ external_id: "ES123:321 block 192.168.0.1-50" }
```

Would it be prefarrable that instead of creating a single Kenna asset we instead create one asset per each individual IP address? So in this example we would create 50 Kenna assets like so:
```ruby
{ ip: "192.168.0.1" }
{ ip: "192.168.0.2" }
{ ip: "192.168.0.3" }
...
{ ip: "192.168.0.48" }
{ ip: "192.168.0.49" }
{ ip: "192.168.0.50" }
```

### DIFF
--- a/tasks/connectors/edgescan/lib/edgescan_asset.rb
+++ b/tasks/connectors/edgescan/lib/edgescan_asset.rb
@@ -23,6 +23,10 @@ module Kenna
           "#{data['name']} (ES#{id})"
         end
 
+        def application?
+          data["type"] == "app"
+        end
+
         def location_specifiers
           @location_specifiers ||= @data["location_specifiers"].map do |specifier|
             EdgescanLocationSpecifier.new(self, specifier)
--- a/tasks/connectors/edgescan/lib/edgescan_definition.rb
+++ b/tasks/connectors/edgescan/lib/edgescan_definition.rb
@@ -12,7 +12,7 @@ module Kenna
 
         def to_kenna_definition
           {
-            "scanner_type" => "Edgescan",
+            "scanner_type" => scanner_type,
             "scanner_identifier" => data["id"],
             "name" => data["name"],
             "description" => data["description_src"],
@@ -23,6 +23,10 @@ module Kenna
         end
 
         private
+
+        def scanner_type
+          data["layer"] == 7 ? "EdgescanApp" : "EdgescanNet"
+        end
 
         def cves
           data["cves"].empty? ? nil : data["cves"].join(",")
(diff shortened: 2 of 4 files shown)


lindambrown:
1. We DO want application_identifier if the main identifier is URL
2. We can only store one ip address. External id is fine but most customers build searches based on hostname/ip rather than external id. Can you add the first IP addess perhaps? if there is no hostname?

edblaz:
> 1. We DO want application_identifier if the main identifier is URL
> 2. We can only store one ip address. External id is fine but most customers build searches based on hostname/ip rather than external id. Can you add the first IP addess perhaps? if there is no hostname?

Thank you for such a quick response! That's perfect, we'll start working on those two points now

edblaz:
Hey @lindambrown! We just made all the changes discussed, here's a quick summary.

#### Commit https://github.com/KennaSecurity/toolkit/pull/241/commits/e0f8e457ae4eac6afe4a2993805a021fada56933:
- `scanner_type` will now be either `EdgescanNet` or `EdgescanApp` depending on the layer of the vulnerability.

#### Commit https://github.com/KennaSecurity/toolkit/pull/241/commits/6ee4c2edd111fb9ec6dbd251ab9993a4bb6bdd69:
- Only application assets with a URL locator will set the `application_identifier`.
- Edgescan assets with IP or CIDR blocks will now populate their respective Kenna asset's `ip_address` with the first address in their block. This will ensure that from now on all assets uploaded from Edgescan will have at least one of these three locators: `ip`, `hostname`, `url`.

Do not hesitate to let us know if you have any gripes or questions

lindambrown (approved):

PR #242, opened 2021-11-02
Add a helper method to fail tasks with a non-zero status code

bschmeck:
As we move to a world where the toolkit is run in hosted containers, having a non-zero status code will make failures easier to detect.  Passing a status value to [`exit`](https://ruby-doc.org/core-2.6.4/SystemExit.html) causes the Ruby process to terminate with the given value as it's status code.

### DIFF
--- a/lib/helpers.rb
+++ b/lib/helpers.rb
@@ -74,6 +74,11 @@ module Kenna
         end
       end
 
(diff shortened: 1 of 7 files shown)


jgamblin (approved):
LGTM!

PR #243, opened 2021-11-03
Exit with non-zero status code in a few more places

bschmeck:
After #242, I found a few more places that can call `fail_task` when exiting.  They were missed initially because they were calling `return` instead of `exit`.

Additionally, early on in the script we will break if given improperly formatted command line arguments, we should exit with a non-zero status code there, also.


jgamblin (approved):
LGTM

PR #258, opened 2021-11-12
Added remove html tags method

fshaiksecurview:


tasks/connectors/qualys_was/README.md
```diff
@@ -1,41 +1,41 @@
...
 - Post the file to Kenna if API Key and Connector ID are provided
 
 ## Things you will need
 
-- Aqua Username (Required)
-- Aqua Password (Required)
-- Aqua Console Address (Required)
+- QualysWas Username (Required)
+- QualysWas Password (Required)
+- QualysWas Console Address (Required)
```
lindambrown, line 14: these should match the parameter names in the code. Port is missing too. Also, what is "container data"? Based on the description it is a list of severities? that doesn't seem right either. is there a numerical filter allowed since Qualys scores 1-5?

kdavesecurview: i thik we should change this to. `qualys_was_domain`  as this will be used as prefix/ domain of the api which is `qualysapi.qg3.apps.qualys.com` or 

we can remove  this and hardcode the api urls' as it is now  https://github.com/KennaSecurity/toolkit/blob/a57d942d04c38bcf01d6fbc8d828e8106202eed8/tasks/connectors/qualys_was/lib/qualys_was_helper.rb#L17 


i prefer making it dynamic, what do you suggest

lindambrown: yes that has to be dynamic because customers could be in any of 8+ domains

lindambrown: and i like the name change

kdavesecurview: name change ?

lindambrown (approved):

PR #264, opened 2021-11-16
Proposed documentation changes

jscalvo:


dbrothersKenna:
The doc looks good to me, but wondering if we should break the "developer's guide" info to a separate page, and leave the initial readme as strictly an operational page.

@lindambrown  Thoughts?

lindambrown:
yeah seem like we need a dev_guide.md that we can point to from other places. @dbrothersKenna is that what you were thinking?

jscalvo:
I could move it to another page.
Would you like a link from the README.md page?
Would you like the JSON KDI format reference inlined in this dev_guide.md?

dbrothersKenna:
Yep.  That's exactly what I was thinking.  Keep the original readme page from getting too long and many won't be doing dev but just using.
I think that's a good idea to put the JSON on that page as well.

jscalvo:
@lindambrown @dbrothersKenna There is this wiki page https://github.com/KennaSecurity/toolkit/wiki/Developers, seems it was a beginning for the development documentation but didn't advance.
What do you think about working there? I'd rather use the wiki.

lindambrown:
Sounds good!

jscalvo:
This will be done in the wiki. I'll submit another PR with the changes to the `_sample_task` folder

PR #270, opened 2021-11-18
This commit adds a new AppScan on Cloud connector.

jscalvo:
There are some TO-DO's to resolve before considering it  OK for testing.
We need a valid license in order to get all data fields and correctly map the locator fields.

### DIFF
--- a/tasks/connectors/appscan_cloud/appscan_cloud_task.rb
+++ b/tasks/connectors/appscan_cloud/appscan_cloud_task.rb
@@ -1,0 +1,242 @@
+# frozen_string_literal: true
+
+require_relative "lib/appscan_cloud_client"
+module Kenna
+  module Toolkit
+    class AppScanCloudTask < Kenna::Toolkit::BaseTask
+      def self.metadata
+        {
+          id: "appscan_cloud",
+          name: "AppScan on Cloud",
+          description: "Pulls assets and vulnerabilities from AppScan on Cloud",
+          options: [
+            { name: "appscan_cloud_key_id",
+              type: "api_key",
+              required: true,
+              default: nil,
+              description: "AppScan Cloud API Key ID" },
+            { name: "appscan_cloud_key_secret",
+              type: "api_key",
+              required: true,
+              default: nil,
+              description: "AppScan Cloud API Key Secret" },
+            { name: "appscan_cloud_applications",
+              type: "string",
+              required: true,
+              default: nil,
+              description: "A list of AppScan on Cloud Applications ID's (comma separated)" },
+            { name: "appscan_cloud_severities",
+              type: "string",
+              required: false,
+              default: nil,
+              description: "A list of ['Undetermined', 'Informational', 'Low', 'Medium', 'High', 'Critical'] (comma separated)" },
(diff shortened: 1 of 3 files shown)


tasks/connectors/appscan_cloud/appscan_cloud_task.rb
```diff
@@ -0,0 +1,242 @@
...
+              finding = extract_finding(issue)
+              definition = extract_definition(issue)
+
+              create_kdi_asset_finding(asset, finding)
+              create_kdi_vuln_def(definition)
+            end
+            print_good("Processed #{[pos + @batch_size, total_issues].min} of #{total_issues} issues for application ##{application_names[application_id]}.")
+            kdi_upload(@output_directory, "appscan_cloud_application_#{application_id}_report_#{pos}.json", @kenna_connector_id, @kenna_api_host, @kenna_api_key, @skip_autoclose, @retries, @kdi_version)
+            pos += @batch_size
+            break if pos >= total_issues
```
lindambrown, line 84: would this be better listed as page size instead of batch size?  @jscalvo

jscalvo: The parameter name is also `batch_size`. Do you mean renaming all references including the parameter name? I used this parameter name in all the other connectors, so must change everywhere. No problem if we define a standardized name for this commonly used param I'd agree it must be changed. Should we use page_size?

lindambrown: i was using batch size for "how much we collect before we send to Kenna" and page size to be "how many rows do we request from the vendor at at time" I don't remember seeing issues with previous naming.

jscalvo: I think that I was using the same parameter for both in my loops. Some don't use the page param, for instance when a report is downloaded with all the issues. Bugcrowd, Burp, JFrog are using batch_size for batches and pages. For me is a bit overkilling using 2 parameters. Maybe with Bugcrowd has more sense because page size is limited to 100, a really small number, so maybe you want to batch by 200 or 500. Anyway, I feel confusing for the end-users having to explain the differences between 2 similar parameters. As I'm new to the team I'm not aware of all the problems you had before, so, if the recommendation is always to use 2 params when possible, I'll follow it.

lindambrown: @dbrothersKenna @bschmeck what do you guys think about this? I have traditionally kept the names different. Page size for how we pull from the vendor and batch size for how often we send to Kenna. I have had at least one connector where I used both for different reasons.

lindambrown: @jscalvo please change this to page size as there have been cases where both are needed. too many calls to either the host api or kenna api can slow things down and there have been cases when page size (how many you get from the vendor at a time) and batch size (how many to have before sending to kenna) differs.

jscalvo: OK, no problem

lindambrown (approved):

PR #271, opened 2021-11-19
Bugcrowd changes requested by Linda & Venkata because the description field do…

jscalvo:
…esn't belong to the vuln definition. Moved to additional fields.

### DIFF
--- a/tasks/connectors/bugcrowd/bugcrowd_task.rb
+++ b/tasks/connectors/bugcrowd/bugcrowd_task.rb
@@ -181,7 +181,6 @@ module Kenna
       def extract_definition(issue)
         {
           "name" => issue["attributes"]["vrt_id"],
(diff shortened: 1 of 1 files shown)


tasks/connectors/bugcrowd/bugcrowd_task.rb
```diff
@@ -181,21 +181,22 @@ def extract_finding(issue)
       def extract_definition(issue)
         {
           "name" => issue["attributes"]["vrt_id"],
-          "description" => CGI.escapeHTML("#{Sanitize.fragment(issue['attributes']['title'])}\n\n#{Sanitize.fragment(issue['attributes']['description'])}"),
           "solution" => issue["attributes"]["remediation_advice"],
```
lindambrown, line 184: @jscalvo I think the solution need to move too unless you can see in the data that it is common across items. from the data we looked at it was unique to the finding. This would leave the vuln_def with only the name and scanner_type

lindambrown (approved):

PR #275, opened 2021-11-25
Remove old sample_task_objects folder and add _sample_task as example…

jscalvo:
… based on current Burp code. This new folder name is referenced in the new documentation on GitHub wiki.


tasks/connectors/_sample_task/sample_task.rb
```diff
@@ -0,0 +1,174 @@
...
+      end
+
+      SEVERITY_VALUE = {
+        "info" => 0,
+        "low" => 3,
+        "medium" => 6,
+        "high" => 10
+      }.freeze
+
+      def extract_asset(issue)
```
lindambrown, line 131: @jscalvo  i think there was benefit in the old sample to having all the possible values with certain things commented out. either that or there we should have 2 with one having all kdi fields and this one being a specific "findings" sample.

jscalvo: Ok, I Agree, let me correct this.

lindambrown (approved):

lindambrown:
@jscalvo there are a few fields missing for findings like due date. Can you add them back in as comments?

PR #282, opened 2021-11-30
Native Ruby Test

jgamblin:
Attempt to run the application natively in ruby.


kenna-bmcdevitt (approved):
👍

bschmeck:
This PR adds `active_support` as a dependency, but in an [earlier PR](https://github.com/KennaSecurity/toolkit/pull/237) we brought over individual methods from that gem instead of adding the gem itself.

I don't have strong feelings either way, but if we are going to add the gem, we should remove those individual methods we brought over.

cc: @lindambrown

jscalvo:
Ben, I think there is no need to add active_support, the problem is with the qualys_was task. As Linda ask me to do some changes I'm going to remove the dependencies and also remove active_support after. Don't worry.

PR #301, opened 2021-12-07
Add activesupport dependency

jscalvo:
qualys_was introduced a dependency on activesupport to parse XML. We choose activesupport over nokogiri because all the useful extensions are well known by most Rails users. Also, the toolkit started to add more and more of those extensions as monkey patches and we don't want to reinvent the wheel.


spec/extensions/hash_spec.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+require_relative "../rspec_helper"
+
+RSpec.describe "Hash" do
```
bschmeck, line 5: Most of the tests here are for methods that are now implemented by ActiveSupport.  I don't think we need to include tests for the functionality we're getting from gems.

spec/extensions/presence_spec.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
-require "rspec_helper"
-require_relative "../initialize/presence"
+require_relative "../rspec_helper"
```
bschmeck, line 3: This spec file is testing the presence monkeypatches, which were removed.  I think this spec can be removed, too.

bschmeck (dismissed):
Adding the ActiveSupport dependency looks correct to me.  I would suggest removing the spec files that test methods provide by ActiveSupport, though.  (It looks like there is also a spec file for the monkeypatches to `Enumerable` that can be removed.)

spec/extensions/presence_spec.rb
jscalvo: Yeah, I leave it there only for the reviewer but now that's approved I'll submit the removal

lindambrown (approved):

PR #350, opened 2022-01-31
Insight AppSec conector

mukhaSS:
Need to check in Kenna

### DIFF
--- a/tasks/connectors/insight_appsec/insight_appsec_task.rb
+++ b/tasks/connectors/insight_appsec/insight_appsec_task.rb
@@ -1,0 +1,219 @@
+# frozen_string_literal: true
+
+require_relative "lib/insight_appsec_client"
+require "pry"
+
+module Kenna
+  module Toolkit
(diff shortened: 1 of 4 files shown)


tasks/connectors/insight_appsec/insight_appsec_task.rb
```diff
@@ -0,0 +1,219 @@
+# frozen_string_literal: true
+
+require_relative "lib/insight_appsec_client"
+require "pry"
```
jscalvo, line 4: Don't forget to remove the require "pry". I suspect you are using RubyMine, why don't you use RM debugger?

mukhaSS: shure) this habit came from another code editor 😅

jscalvo (changes requested):

lindambrown (approved):

lindambrown (approved):

PR #371, opened 2022-03-01
Add verbosity to API exception logging to help clients/support

sjbridge:
This change adds more details to logging about API call exceptions. This will help clients know whether the failures they see logged are in their scanner or Kenna.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -331,7 +331,7 @@ Style/StringLiterals:
   EnforcedStyle: double_quotes
 
 Layout/SpaceAroundMethodCallOperator:
-  Enabled: true  
+  Enabled: true
 
 Lint/RaiseException:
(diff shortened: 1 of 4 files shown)


BlakeHendersonKS (dismissed):
After chatting with Sam, this additional logging should provide some valuable insight for our clients trying to debug issues going forward.

jscalvo:
I'm not sure about this PR. I'll check with the rest of the team

jscalvo:
I think we can change it this way:
1. Add `debug?` check
2. Add a new method to check if it's running on the cloud, for example, `:running_on_kenna_hosted?`
The method probably could check if some environment variable is present
3. `print_request_error_retails if debug? && !running_on_kenna_hosted?`

richauha (approved):

PR #383, opened 2022-04-01
Con 1165 netsparker wildcard

annapoltieva:
**JIRA** [CON-1165](https://kennasecurity.atlassian.net/browse/CON-1165)

**Problem**
Lowes has a large number of scheduled ids and it is hard for them to find them.

**Solution**
Turn netsparker_scheduled_ids argument into an optional one and pull all the scheduled ids from the API if this argument is not given or set to *


mukhansk (approved):

tasks/connectors/netsparker/lib/netsparker_client.rb
```diff
@@ -23,6 +23,23 @@ def get_last_scan_vulnerabilities(schedule_id)
...
+          schedule_ids = []
+          loop do
+            response = http_get(list_scheduled_url(page), @headers)
+            scheduled_scan_result = JSON.parse(response)
+            schedule_ids.push(*scheduled_scan_result.fetch("List").map { |scan| scan.fetch("Id") }.uniq)
+            break if scheduled_scan_result["IsLastPage"]
+
+            page += 1
+          end
+          schedule_ids.uniq
```
jscalvo, line 37: Why is `uniq` called again? Is called in the loop also.

tasks/connectors/netsparker/lib/netsparker_client.rb
```diff
@@ -23,6 +23,23 @@ def get_last_scan_vulnerabilities(schedule_id)
...
+        rescue KeyError
+          print_error "There are no scheduled scans"
+          exit 1
```
jscalvo, line 40: There is a helper for above 2 lines:
```ruby
fail_task "There are no scheduled scans"
```

tasks/connectors/netsparker/lib/netsparker_client.rb
```diff
@@ -23,6 +23,23 @@ def get_last_scan_vulnerabilities(schedule_id)
           JSON.parse(response)
         end
 
+        def receive_all_schedule_ids
```
jscalvo, line 26: I'd change the method name to something like `get_all_schedule_ids` or `retrieve_all_scheduled_ids`.

jscalvo (changes requested):

tasks/connectors/netsparker/lib/netsparker_client.rb
annapoltieva: My bad. It should just be called once outside of the loop

jscalvo (approved):

PR #388, opened 2022-04-13
Move http_get(list_scheduled_url(page), @headers) from loop

mukhansk:
[CON-1224](https://kennasecurity.atlassian.net/browse/CON-1224)

### DIFF
--- a/tasks/connectors/netsparker/lib/netsparker_client.rb
+++ b/tasks/connectors/netsparker/lib/netsparker_client.rb
@@ -5,6 +5,7 @@ module Kenna
     module Netsparker
       class NetsparkerClient
         HOST = "https://www.netsparkercloud.com"
+
         def initialize(user_id, token)
           auth_token = Base64.strict_encode64("#{user_id}:#{token}")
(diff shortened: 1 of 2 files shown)


tasks/connectors/netsparker/lib/netsparker_client.rb
```diff
@@ -44,14 +47,16 @@ def retrieve_all_scheduled_ids
         def get_last_scan_id(schedule_id)
           found = nil
           page = 1
+          response = http_get(list_scheduled_url(page), @headers)
+          scheduled_scan_result = JSON.parse(response)
+
```
annapoltieva, line 52: That's better, but we still fetch all scheduled ids for every entry in @scheduled_ids

annapoltieva (changes requested):
We fetch all scheduled ids before line 70 in netsparker_task.rb and pass them as a parameter to get_last_scan_id so that we don't have to re-fetch them every time

annapoltieva (approved):

annapoltieva (approved):

PR #393, opened 2022-04-20
Con 1261

annapoltieva:
CON-1261

## Problem
Ruby 2.6 is reaching EOL and we need to upgrade.

## Solution
Upgrade tp 3.1.2


lib/api/client.rb
```diff
@@ -144,7 +144,7 @@ def upload_to_connector(connector_id, filepath, run_now = true, max_retries = 3,
           query_response = RestClient::Request.execute(
             method: :post,
             url: connector_endpoint,
-            headers: headers,
```
charles-sturgis-kenna, line 147: RestClient's version didn't bump - how do we have a different signature?

charles-sturgis-kenna: Are these in response to a rubocop? Also, note the trailing colon - we've got `headers:,` instead of `headers,`

charles-sturgis-kenna (changes requested):
Let's look at these changed signatures

lib/api/client.rb
mukhansk: Yes rubocop say that now we need write ```headers:,``` instead of ```headers: headers,```

charles-sturgis-kenna (approved):

PR #403, opened 2022-05-24
Update Edgescan Connector

ShayVD:
# Context

Clients require the ability to import vulnerabilities based on layer type, `application` and `network`.
Bug also found when attempting to create Kenna findings.

# Changes

- Added 2 flags `include_network_vulnerabilities`, and `include_application_vulnerabilities` to allow the separation of vulnerabilities when importing. Both set to `true` by default.
- Fixed bug where Kenna findings were not being created when `create_findings` flag was set to `true`.

### DIFF
--- a/tasks/connectors/edgescan/edgescan.rb
+++ b/tasks/connectors/edgescan/edgescan.rb
@@ -69,7 +69,11 @@ module Kenna
         edgescan_api.fetch_in_batches do |edgescan_assets, edgescan_definitions|
           edgescan_assets.each do |edgescan_asset|
             kenna_api.add_assets(edgescan_asset)
-            kenna_api.add_vulnerabilities(edgescan_asset.vulnerabilities)
+            if @options[:create_findings]
+              kenna_api.add_vulnerabilities(edgescan_asset.vulnerabilities)
+            else
+              kenna_api.add_findings(edgescan_asset.vulnerabilities)
+            end
           end
 
           kenna_api.add_definitions(edgescan_definitions)
--- a/tasks/connectors/edgescan/edgescan.rb
+++ b/tasks/connectors/edgescan/edgescan.rb
@@ -55,7 +55,17 @@ module Kenna
               type: "boolean",
               required: false,
               default: false,
-              description: "The task will create findings, instead of vulnerabilities" }
+              description: "The task will create findings, instead of vulnerabilities" },
+            { name: "include_network_vulnerabilities",
+              type: "boolean",
+              required: false,
+              default: true,
+              description: "The task will include network layer vulnerabilities" },
+            { name: "include_application_vulnerabilities",
+              type: "boolean",
+              required: false,
+              default: true,
+              description: "The task will include application layer vulnerabilities" }
           ]
         }
       end
--- a/tasks/connectors/edgescan/edgescan.rb
+++ b/tasks/connectors/edgescan/edgescan.rb
@@ -81,9 +81,9 @@ module Kenna
             kenna_api.add_assets(edgescan_asset)
             if @options[:include_network_vulnerabilities] || @options[:include_application_vulnerabilities]
               if @options[:create_findings]
-                kenna_api.add_vulnerabilities(edgescan_asset.vulnerabilities)
-              else
                 kenna_api.add_findings(edgescan_asset.vulnerabilities)
+              else
+                kenna_api.add_vulnerabilities(edgescan_asset.vulnerabilities)
               end
             end
           end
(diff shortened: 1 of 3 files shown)


tasks/connectors/edgescan/edgescan.rb
```diff
@@ -69,7 +79,11 @@ def run(opts)
         edgescan_api.fetch_in_batches do |edgescan_assets, edgescan_definitions|
           edgescan_assets.each do |edgescan_asset|
             kenna_api.add_assets(edgescan_asset)
-            kenna_api.add_vulnerabilities(edgescan_asset.vulnerabilities)
+            if @options[:create_findings]
```
whitfieldsdad, line 82: Should the `--create-findings` flag disable the import of vulnerabilities, or should it simply enable the import of findings?

tasks/connectors/edgescan/lib/edgescan_api.rb
```diff
@@ -74,6 +76,18 @@ def base_url
 
           "https://#{@api_host}"
         end
+
+        # Application vulnerabilities are on layer 7
+        # Network vulnerabilities are on every other layer
+        def layer_query_parameters
+          if @include_application_vulnerabilities && @include_network_vulnerabilities
```
whitfieldsdad, line 83: What happens if a user disables the import of both application layer and network layer vulnerabilities?

whitfieldsdad (changes requested):

tasks/connectors/edgescan/edgescan.rb
ShayVD: As far as I can tell edgescan vulnerabilities should be imported as vulnerabilities OR findings.

From the feature request:
"Kenna are updating their data model and have asked us to add a toggle that controls whether to create vulnerabilitys or findings for apps ([the differences are outlined here](https://help.kennasecurity.com/hc/en-us/articles/360026413111-Kenna-Data-Importer-JSON-Connector-))"

tasks/connectors/edgescan/lib/edgescan_api.rb
ShayVD: Fixed in commit dd27498.
If a user disables both application and network layer vulnerabilities then no vulnerabilities will be imported, but assets will still be created.

tasks/connectors/edgescan/edgescan.rb
whitfieldsdad: @annapoltieva: are vulnerabilities and findings mutually exclusive?

annapoltieva: Sorry to leave you hanging, but I still discuss that with the business

tasks/connectors/edgescan/edgescan.rb
```diff
@@ -79,10 +79,12 @@ def run(opts)
         edgescan_api.fetch_in_batches do |edgescan_assets, edgescan_definitions|
           edgescan_assets.each do |edgescan_asset|
             kenna_api.add_assets(edgescan_asset)
-            if @options[:create_findings]
-              kenna_api.add_vulnerabilities(edgescan_asset.vulnerabilities)
-            else
-              kenna_api.add_findings(edgescan_asset.vulnerabilities)
+            if @options[:include_network_vulnerabilities] || @options[:include_application_vulnerabilities]
+              if @options[:create_findings]
```
whitfieldsdad, line 83: This conditional appears to be backwards - looks like a typo to me

ShayVD: So this is saying that if you want to import any edgescan vulnerabilities (network, application, or both), do you want to create Kenna findings or Kenna vulnerabilities from them then.
I believe this is the desired functionality.

tasks/connectors/edgescan/edgescan.rb
annapoltieva: The answer - We definitely want findings OR vulns. Findings is the new model, vulns is old model (for AppSec Only. for VM it will always be vulns).

tasks/connectors/edgescan/edgescan.rb
whitfieldsdad: Oh, this one:

```
if @options[:create_findings]
    kenna_api.add_vulnerabilities(edgescan_asset.vulnerabilities)
else
    kenna_api.add_findings(edgescan_asset.vulnerabilities)
```

This condition says to create vulnerabilities if the create findings flag is set and vice-versa.

ShayVD: Fixed in [3d5efeb](https://github.com/KennaSecurity/toolkit/pull/403/commits/3d5efebc2754ca3214e37519754d284cf2da7ae9)

annapoltieva (approved):
LGTM

mukhansk:
Please run
`bundle update rack`
to Bundler Audit / audit (pull_request) be successful

mukhansk (approved):

PR #404, opened 2022-05-25
Correct Wiz.io documentation

BlakeHendersonKS:
The documentation currently indicates that we're able to pull vulnerabilities detected after a given date, when really we're only able to pull vulnerabilities created after a given date.


annapoltieva (approved):
LGTM

PR #409, opened 2022-06-20
JETS-147 Update how location is provided to assets

ShayVD:
# Context

Some assets being created in Kenna from edgescan data did not have the location (ip/hostname/url) included.
Also, ip address ranges and blocks were being cut short.

# Changes

Include location when creating an asset from a vulnerability.
Include full ip address blocks or ranges.


whitfieldsdad (changes requested):
It looks like the Edgescan connector translates Edgescan's definition of scan ranges into Kenna's definition of individual computers:

See:
- https://github.com/KennaSecurity/toolkit/blob/main/tasks/connectors/edgescan/lib/edgescan_api.rb#L49-L51
- https://github.com/KennaSecurity/toolkit/blob/main/tasks/connectors/edgescan/lib/edgescan_asset.rb#L41-L54
- https://github.com/KennaSecurity/toolkit/blob/main/tasks/connectors/edgescan/lib/edgescan_location_specifier.rb#L26-L38

Should we instead be translating hosts into hosts?

Here's an example of what an Edgescan asset looks like:

```
{
  "id": █████,
  "asset_status": "validating",
  "authenticated": █████,
  "active_license": {
    "id": █████,
    "name": "█████",
    "license_type_id": █████,
    "license_type_name": "█████",
    "asset_id": null,
    "order_id": █████,
    "start_date": "█████",
    "end_date": "█████",
    "expired": false,
    "status": null
  },
  "blocked_status": "unblocked",
  "created_at": "█████",
  "current_assessment": {
    "id": █████,
    "type": "scheduled",
    "status": "validating",
    "created_at": "█████"
  },
  "host_count": █████,
  "hostname": "█████, █████, █████, █████, █████",
  "last_assessment_date": "█████,
  "last_host_scan": "█████",
  "linked_assets": [],
  "location_specifiers": [
    {
      "id": █████,
      "location": "█████",
      "location_type": "cidr"
    },
    {
      "id": █████,
      "location": "█████",
      "location_type": "ip"
    },
    {
      "id": █████,
      "location": "█████",
      "location_type": "ip"
    },
    {
      "id": █████,
      "location": "█████",
      "location_type": "ip"
    },
    {
      "id": █████,
      "location": "█████",
      "location_type": "ip"
    }
  ],
  "name": "█████",
  "network_access": "█████",
  "next_assessment_date": "█████",
  "pci_enabled": █████,
  "priority": █████,
  "tags": [
    "█████"
  ],
  "targeting_mode": "automatic",
  "type": "net",
  "updated_at": "█████"
}
```

Here's an example of what an Edgescan host looks like:

```
{
  "id": █████,
  "asset_id": █████,
  "location": "█████",
  "hostnames": [],
  "label": null,
  "status": "alive",
  "updated_at": "█████",
  "os_name": "Linux 2.6.X",
  "apis_detected": █████
}
```

Here's an example of what a Kenna asset looks like:

```
{
  "application": █████,
  "asset_groups": [],
  "container": █████,
  "created_at": "█████",
  "database": █████,
  "ec2": █████,
  "external_id": "█████",
  "file": █████,
  "fqdn": █████,
  "hostname": "█████",
  "id": █████,
  "image": █████,
  "inactive_at": "█████",
  "ip_address": "█████",
  "ipv6": █████,
  "last_booted_at": █████,
  "last_seen_time": "█████",
  "locator": "█████",
  "mac_address": "█████",
  "netbios": █████,
  "network_ports": [],
  "notes": █████,
  "operating_system": "█████",
  "owner": █████,
  "primary_locator": "mac_address",
  "priority": █████,
  "risk_meter_score": █████,
  "status": "█████",
  "status_set_manually": █████,
  "tags": [],
  "url": █████,
  "urls": {
    "vulnerabilities": "api.kennasecurity.com/assets/█████/vulnerabilities"
  },
  "vulnerabilities_count": █████
}
```

If I were writing the Edgescan connector, I'd to use the `/api/v1/hosts` endpoint rather than the `/api/v1/assets` endpoint.

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
```diff
@@ -50,7 +53,19 @@ def to_kenna_finding
         end
 
         def to_corresponding_kenna_asset
-          { "external_id" => external_asset_id, "tags" => asset.tags }
+          base = { "external_id" => external_asset_id, "tags" => asset.tags }
```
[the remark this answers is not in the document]
ShayVD: @whitfieldsdad So what's not mentioned in the documentation there as far as I can see, is that in edgescan an `Asset` can have many `Location Specifiers`, which could be a `url`, `hostname`, `ip`, `cidr`, or `block`. What the edgescan connector is doing is translating each `Location Specifier` into a Kenna asset.

Sometimes a vulnerability is found in edgescan which doesn't have an exact matching  `Location Specifier`. In this case a Kenna asset will be created, using the location where the vulnerability was found, so the vulnerability can be linked to it.

Using `Hosts` wouldn't capture everything from edgescan's side either. Not every edgescan `Asset` or `Location Specifier` has a matching host. The `Host` information is auto-generated by the ASM scans we do. Where as an `Asset` is added by a person.

tasks/connectors/edgescan/lib/edgescan_location_specifier.rb
```diff
@@ -28,9 +28,7 @@ def to_kenna_asset
 
           base["url"] = location if type == "url"
           base["hostname"] = location if type == "hostname"
-          base["ip_address"] = location if type == "ip"
-          base["ip_address"] = location.split("-").first if type == "block"
-          base["ip_address"] = location.split("/").first if type == "cidr"
+          base["ip_address"] = location if ["ip", "block", "cidr"].include?(type)
```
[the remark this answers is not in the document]
whitfieldsdad: In Kenna, an asset is a single computer, but in EdgeScan, an asset represents a range of computers (e.g. all of the hosts in one or more subnets).

Are we saying that computers can be assigned blocks of IP addresses (e.g. all 254 IP address in a /24 block), or are they generally assigned 1 or 2?

whitfieldsdad, line 31: Since we're now translating Edgescan hosts (and/or vulnerabilities) into Kenna assets, and not translating Edgescan assets into Kenna assets, do we still need this file?

I went through the references, and it looks like this code isn't used for anything.

tasks/connectors/edgescan/lib/edgescan_asset.rb
```diff
@@ -33,34 +34,34 @@ def location_specifiers
...
+
+          host = hosts.find { |h| h.hostnames.each { |hostname| location.include?(hostname) } }
+          return host unless host.nil?
+
+          nil
         end
 
-        def vulnerabilities_without_location_specifiers_as_kenna_assets
-          vulnerabilities.reject(&:matching_location_specifier).map(&:to_corresponding_kenna_asset)
+        def find_location_specifier(specifier_id, location)
```
whitfieldsdad, line 62: IMO this code probably isn't needed as we should always be able to find a host in Edgescan if there's a vulnerability related to it - right?

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
```diff
@@ -50,11 +53,21 @@ def to_kenna_finding
...
+          if ip_address?
+            base["ip"] = location
+          elsif url?
+            base["url"] = location
+          else
```
whitfieldsdad, line 61: Is there a way to retrieve _both_ the IP address and the hostname associated with a particular system?

i.e.

```
{
  "ip_address": "10.15.1.32",
  "hostname": example.pizza,
  "tags": ["YOLO"],
}
```

tasks/connectors/edgescan/lib/edgescan_asset.rb
```diff
@@ -33,34 +34,34 @@ def location_specifiers
...
-        # - Go through the vulnerabilites and if some of them don't have a matching Edgescan
-        #   location specifier create Kenna assets for them
+        # - Create Kenna assets based on Edgescan hosts
+        # - If a matching host is not found an asset will be created from the vulnerability data
         def to_kenna_assets
-          location_specifiers_as_kenna_assets +
-            vulnerabilities_without_location_specifiers_as_kenna_assets
+          vulnerabilities.each_with_object([]) do |vulnerability, assets|
+            host = find_host(vulnerability.location)
+            asset = host ? host.to_corresponding_kenna_asset : vulnerability.to_corresponding_kenna_asset
```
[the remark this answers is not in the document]
ShayVD: Hosts and Vulnerabilities may not always have the same location. A Host's location is always an IP, whereas a Vulnerability's location could be a hostname, URL, or IP.

A Host may not be present for all Vulnerabilities in edgescan either. Hosts are auto-created when an ASM scan is run. But there are other scans (VM scans) where a Host object is not created. In this case we would need the fallback.

It'll be fine for you as your may just be using ASM scanning, but would be required for other clients.

tasks/connectors/edgescan/lib/edgescan_asset.rb
```diff
@@ -33,34 +34,34 @@ def location_specifiers
...
+            assets << asset
+          end
         end
 
         private
 
-        def location_specifiers_as_kenna_assets
-          location_specifiers.map(&:to_kenna_asset)
+        def find_host(location)
```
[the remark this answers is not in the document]
ShayVD: A Host's location will always be an IP. A Vulnerability's location could potentially be "https://test.example.com/some/page.html", for example.

tasks/connectors/edgescan/lib/edgescan_asset.rb
ShayVD: For you, yes you should always be able to find a Host since you're running ASM scans, for other clients running VM scans then a Host won't be available. 

Although, we're not using location specifiers anymore anyway so, removed in f69b7fc.

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
ShayVD: Not when creating an asset from the Vulnerability no. Since this is used when no Host is available we don't have the IP and hostname available.

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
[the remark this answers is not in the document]
whitfieldsdad: That makes sense, thanks!

I think that the best bet for simplifying the logic here would be to avoid using the strategy pattern and instead have one only parser per output object type (i.e. one function for creating Kenna assets, one for creating vulnerabilities, and one for creating findings).

You could then pass in any additional information if it's available (e.g. hostnames from host objects).

I don't know Ruby very well, but in Python, you could do something like this:

```python
def parse_host(asset: dict, vulnerability: dict, host: Optional[dict] = None) -> dict:
  hostname = None
  if host:    
    hostnames = host['hostnames']
    if hostnames:
      hostname = hostnames[0]

return {
    'ip_address': asset['locator'],
    'tags': asset['tags'],
    'hostname': hostname
  }
```

tasks/connectors/edgescan/lib/edgescan_asset.rb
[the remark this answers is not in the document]
whitfieldsdad: Makes sense - it might be a good idea to add nullable fields on the backend for things like URLs, hostnames, IP addresses, etc., but that's out-of-scope for this discussion - might make things easier for you and customers though.

whitfieldsdad (changes requested):

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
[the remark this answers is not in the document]
whitfieldsdad: AFAIK this change should basically cut your code in half

tasks/connectors/edgescan/lib/edgescan_asset.rb
```diff
@@ -33,34 +34,34 @@ def location_specifiers
...
+          host = hosts.find { |h| h.location == location }
+          return host unless host.nil?
+
+          host = hosts.find { |h| h.hostnames.each { |hostname| location.include?(hostname) } }
+          return host unless host.nil?
+
+          nil
```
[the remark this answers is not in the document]
ShayVD: Fixed in [55031b4](https://github.com/KennaSecurity/toolkit/pull/409/commits/55031b4c31b0f90f022e2f7dd3bdd00c048ba55b)

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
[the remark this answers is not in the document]
ShayVD: Fixed in [55031b4](https://github.com/KennaSecurity/toolkit/pull/409/commits/55031b4c31b0f90f022e2f7dd3bdd00c048ba55b)

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
```diff
@@ -49,12 +45,33 @@ def to_kenna_finding
...
-        def to_corresponding_kenna_asset
-          { "external_id" => external_asset_id, "tags" => asset.tags }
+        def to_kenna_asset
+          base = {
+            "external_id" => external_asset_id,
+            "tags" => asset["tags"],
+            "application" => asset_application_id
+          }
+          if ip_address?
+            base["ip"] = location
```
whitfieldsdad, line 55: The IP address has to be provided in a field named `ip_address` rather than `ip` - see the asset section of the KDI v2 spec.: https://help.kennasecurity.com/hc/en-us/articles/360026413111-Kenna-Data-Importer-JSON-Connector-

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
```diff
@@ -49,12 +45,33 @@ def to_kenna_finding
...
+          elsif url?
+            base["url"] = location
+          else
+            base["hostname"] = location
+          end
+          if host
+            base["ip"] = host["location"]
```
whitfieldsdad, line 62: The IP address has to be provided in a field named ip_address rather than ip - see the asset section of the KDI v2 spec.: https://help.kennasecurity.com/hc/en-us/articles/360026413111-Kenna-Data-Importer-JSON-Connector-

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
```diff
@@ -49,12 +45,33 @@ def to_kenna_finding
...
+            base["hostname"] = host["hostnames"]&.first
+            base["os"] = host["os_name"]
```
whitfieldsdad, line 64: `os` is intended to mark the OS type not the OS version - to specify the OS version, use the `os_version` property: https://help.kennasecurity.com/hc/en-us/articles/360026413111-Kenna-Data-Importer-JSON-Connector-

whitfieldsdad (changes requested):

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
ShayVD: Fixed in 34602fe

tasks/connectors/edgescan/edgescan.rb
```diff
@@ -76,15 +74,13 @@ def run(opts)
...
-          edgescan_assets.each do |edgescan_asset|
-            kenna_api.add_assets(edgescan_asset)
-            if @options[:include_network_vulnerabilities] || @options[:include_application_vulnerabilities]
-              if @options[:create_findings]
-                kenna_api.add_findings(edgescan_asset.vulnerabilities)
-              else
-                kenna_api.add_vulnerabilities(edgescan_asset.vulnerabilities)
-              end
+        edgescan_api.fetch_in_batches do |edgescan_vulnerabilities, edgescan_definitions|
+          kenna_api.add_assets(edgescan_vulnerabilities.map(&:to_kenna_asset))
```
whitfieldsdad, line 78: This looks perfect! 👏🏻

whitfieldsdad (approved):
🚢

whitfieldsdad:
Hey, @jscalvo, @annapoltieva, @mukhansk - I work at Great-West Life, and have been working with @ShayVD from Edgescan on updates to the Edgescan connector.

The scope of this change is to update how the Edgescan connector translates hosts from Edgescan's perspective into hosts from Kenna's perspective.

The change looks good to me - could you take a look? 

Here is where vulnerabilities identified by Edgescan are translated into assets: https://github.com/KennaSecurity/toolkit/pull/409/files#diff-c39cc6248a94cbf1783fe81377f1faf47c02dd0959a0eefa1e99fa2f44bdabffR48-R67

image

mukhansk:
Hey 👋 )
I think the changes look great. 👍 
I have only one clarification. Did you have had the ability to check this on KENNA (did the files upload successfully on KENNA and whether it is displayed correctly)?

ShayVD:
Hey @mukhansk, yes ituploads and displays correctly in Kenna.

Asset:

image

Vulnerability:

image

mukhansk:
One more clarification.
Did you create a mapping file for edgescan connector and kenna during development?

ShayVD:
@mukhansk Something like this?

image

tasks/connectors/edgescan/README.md
```diff
@@ -33,6 +42,76 @@ These are some quick examples:
 | kenna_api_host     | false    | Kenna API hostname                                                           | api.us.kennasecurity.com |
 | output_directory   | false    | The task will write JSON files here (path is relative to the base directory) | output/edgescan          |
 | create_findings    | false    | The task will create findings, instead of vulnerabilities                    | false                    |
+| network_vulns      | false    | The task will include network layer vulnerabilities                          | true                     |
+| application_vulns  | false    | The task will include application layer vulnerabilities                      | true                     |
+
+## Data Mappings
```
whitfieldsdad, line 48: Wow, this section looks great!

tasks/connectors/edgescan/README.md
```diff
@@ -33,6 +42,76 @@ These are some quick examples:
...
+
+Edgescan assets do not map directly to Kennna assets due to Edgescan assets being more flexible in their definition.
+Edgescan location specifiers and hosts are more like Kenna assets. Location specifiers define the location and hosts hold extra information.
+Not all location specifiers have a host, and not all vulnerabilities have a directly related location specifier.
```
whitfieldsdad, line 52: Why don't vulnerabilities have a location? If we know the system's hostname or IP address, don't we know the vulnerability's location?

tasks/connectors/edgescan/README.md
```diff
@@ -33,6 +42,76 @@ These are some quick examples:
 | kenna_api_host     | false    | Kenna API hostname                                                           | api.us.kennasecurity.com |
 | output_directory   | false    | The task will write JSON files here (path is relative to the base directory) | output/edgescan          |
 | create_findings    | false    | The task will create findings, instead of vulnerabilities                    | false                    |
+| network_vulns      | false    | The task will include network layer vulnerabilities                          | true                     |
+| application_vulns  | false    | The task will include application layer vulnerabilities                      | true                     |
+
+## Data Mappings
+
+Edgescan assets do not map directly to Kennna assets due to Edgescan assets being more flexible in their definition.
```
whitfieldsdad, line 50: I think that there's a misunderstanding surrounding the differences between Kenna and Edgescan's respective data models.

In Kenna, an asset is a host (i.e. a single computer).

In Edgescan, an asset is a group of one or more hosts which are referred to using location specifiers (e.g. using a hostname, IP address, list of IP addresses or a block of IP addresses).

We can think of Edgescan assets as the hosts that Edgescan is configured to scan, and hosts as the systems discovered by a particular scan.

Edgescan assets and Kenna assets don't map 1:1 because they're not the same thing - the connector should be translating Edgescan _hosts_ into Kenna _assets_.

tasks/connectors/edgescan/edgescan.rb
```diff
@@ -71,24 +70,27 @@ def self.metadata
       end
 
       def run(opts)
+        if opts[:option] == "help"
+          print_task_help self.class.metadata[:id]
+          print_good "Returning!"
```
whitfieldsdad, line 75: Is this left over from debugging?

tasks/connectors/edgescan/lib/edgescan_location_specifier.rb
```diff
@@ -24,21 +25,81 @@ def location
...
+          when "block"
+            to_kenna_assets_from_block
+          when "cidr"
+            to_kenna_assets_from_cidr
+          end
+        end
+
+        private
+
+        def to_kenna_assets_from_block
```
whitfieldsdad, line 44: Why are we translating blocks of IP addresses into hosts?

tasks/connectors/edgescan/lib/edgescan_location_specifier.rb
```diff
@@ -24,21 +25,81 @@ def location
...
+            kenna_assets << { **kenna_asset_base, **base, **host_information(new_ip) }
+          end
+        end
 
-          base["url"] = location if type == "url"
-          base["hostname"] = location if type == "hostname"
-          base["ip_address"] = location if type == "ip"
-          base["ip_address"] = location.split("-").first if type == "block"
-          base["ip_address"] = location.split("/").first if type == "cidr"
+        def to_kenna_assets_from_cidr
```
whitfieldsdad, line 55: This will create dummy assets that were never picked up by Edgescan, right?

tasks/connectors/edgescan/README.md
ShayVD: I agree that we should be translating Edgescan hosts into Kenna assets _when possible_. And the connector is making use of host data when it is available. Host data is not present for every location/vulnerability. Host data is auto generated during just one of the many scans that edgescan offers. Therefore we must find alternative ways to group vulnerability information together for when host data is not available.

When creating Kenna assets location specifiers should be used as the edgescan equivalent in my opinion since every host **must** have a corresponding location specifier.

When creating a Kenna asset from a location specifier it will search for a corresponding host. If there is a host then, the host information will be used. If there is no host then the usual location specifier data is used. 

A vulnerability's location may also be used to create a kenna asset in the case that neither location specifier or host data can be found. If there is no host data and the location specifier was deleted, for example.

tasks/connectors/edgescan/edgescan.rb
ShayVD: I added that in as it seems that kenna connector does not properly display the options of a task.

For example if i did `docker run toolkit task=edgescan help` it outputs the same as if I did `docker run help`. It does not display edgescan's options.

tasks/connectors/edgescan/lib/edgescan_location_specifier.rb
ShayVD: It was brought up in the previous meeting that it would be useful to import assets even if there were no vulnerabilities associated with them. If there is no host data and no vulnerability data for a certain location then location specifiers is the only data point that can be used to create a kenna asset.

tasks/connectors/edgescan/lib/edgescan_location_specifier.rb
ShayVD: This will create a kenna asset for each of the locations in a cidr or block ip address range, whether the location has an associated vulnerability or not.

tasks/connectors/edgescan/README.md
```diff
@@ -16,12 +20,17 @@ This task will communicate with the Edgescan and Kenna APIs. In order to do so i
 More in depth details about running the task are available [here](https://github.com/KennaSecurity/toolkit/blob/master/README.md).
 These are some quick examples:
 
-- To print a list of available options: `docker run -it --rm toolkit:latest task=edgescan help`
-- To sync all Edgescan data into Kenna: `docker run -it --rm toolkit:latest task=edgescan edgescan_token='abc' kenna_api_key='abc' kenna_connector_id=123`
+- To print a list of available options: `docker run -it --rm kennasecurity/toolkit:latest task=edgescan option=help`
+- To sync all Edgescan data into Kenna: `docker run -it --rm kennasecurity/toolkit:latest task=edgescan edgescan_token='abc' kenna_api_key='abc' kenna_connector_id=123`
```
whitfieldsdad, line 24: This is fantastic! 🥇

tasks/connectors/edgescan/lib/edgescan_location_specifier.rb
whitfieldsdad: This doesn't make sense to me as we'd be creating assets in Kenna for assets that WERE NOT identified by Edgescan - they're just in a particular IP address range.

tasks/connectors/edgescan/README.md
whitfieldsdad: Yes, every host _must_ have a location specifier, but not every location specifier maps to a host (e.g. a block of 65,000 IP addresses doesn't necessarily map to 65,000 active assets).

tasks/connectors/edgescan/lib/edgescan_location_specifier.rb
whitfieldsdad: Will this result in the Edgescan connector creating assets in Kenna for hosts that have never been identified by Edgescan, and may not even exist on the network?

tasks/connectors/edgescan/README.md
ShayVD: A URL location specifier, `http://demo.testfire.net` for example, will also not have a host associated with it. Vulnerabilities found at this URL may have a location such as `http://demo.testfire.net/error_handling.jsp` or `http://demo.testfire.net/osci.jsp` associated with it. If i reverted the most recent changes we would have an asset per vulnerability, or with the changes the vulnerabilities are grouped under one asset with the original URL. See pictures below.

The fact that some empty assets are also being created seems to be irrelevant as Kenna does not display an asset unless a vulnerability is associated with it. Regardless, I've removed most of the location specifier code although, some of its data will still be used when creating an asset from a vulnerability to keep the grouping described above, and shown below.

I think this grouping is better, but let me know what you think.

Before:
image
image

Currently:
image
image

tasks/connectors/edgescan/lib/edgescan_location_specifier.rb
ShayVD: Removed in ffb6b20.

tasks/connectors/edgescan/lib/edgescan_location_specifier.rb
ShayVD: Removed in ffb6b20.

mukhansk:
Could you add back the assets with zero vulns functionality please? Looks like Kenna does store assets without vulnerabilities.

ShayVD:
> Could you add back the assets with zero vulns functionality please? Looks like Kenna does store assets without vulnerabilities.

Last commit has been undone to add empty assets.

mukhansk (approved):
LGTM

PR #418, opened 2022-07-28
MS Defender TVM updates

venkatawunnava:
Problem
TVM was not bringing in ip address and tags. Customers also reported seeing inactive assets 

Solution
This pr brings in only Active assets and their ip addresses, tags

### DIFF
--- a/tasks/connectors/ms_defender_tvm/ms_defender_tvm.rb
+++ b/tasks/connectors/ms_defender_tvm/ms_defender_tvm.rb
@@ -133,13 +133,16 @@ module Kenna
             machine_id = vuln.fetch("deviceId")
             fqdn = vuln.fetch("deviceName")
 
-            get_ip(machine_id) if not my_array.include? machine_id
+            ip_json_response = get_ip(machine_id) unless my_array.include? machine_id
             my_array |= [machine_id]
-
+            unless ip_json_response.nil?
+               ipadd = ip_json_response["lastIpAddress"]
+               tagss = ip_json_response["machineTags"]
+            end
             # Get the asset details & craft them into a hash
             asset = {
               "external_id" => machine_id,
-              "ip_address" => $ipadd,
+              "ip_address" => ipadd,
               "fqdn" => fqdn,
               "hostname" => fqdn.split(".")[0],
               "os" => vuln.fetch("osPlatform"),
@@ -149,7 +152,7 @@ module Kenna
             }
 
             # Construct tags
-            tags = $tagss
+            tags = tagss
             # tags = []
             # tags << "MSDefenderTvm"
             # tags << "rbacGroup: #{vuln.fetch('rbacGroupName')}" unless vuln.fetch("rbacGroupName").nil?
--- a/tasks/connectors/ms_defender_tvm/ms_defender_tvm.rb
+++ b/tasks/connectors/ms_defender_tvm/ms_defender_tvm.rb
@@ -120,120 +120,121 @@ module Kenna
           vuln_json = vuln_json_response["value"]
 
           vuln_json.each do |vuln|
-              vuln_cve = vuln.fetch("cveId")
-              scanner_id = vuln_cve
-              if vuln_cve.start_with?("CVE")
-                vuln_cve = vuln_cve.strip
-              else
-                vuln_name = vuln_cve
-                vuln_cve = nil
+            vuln_cve = vuln.fetch("cveId")
+            scanner_id = vuln_cve
+            if vuln_cve.start_with?("CVE")
+              vuln_cve = vuln_cve.strip
+            else
+              vuln_name = vuln_cve
+              vuln_cve = nil
+            end
+
+            machine_id = vuln.fetch("deviceId")
+            fqdn = vuln.fetch("deviceName")
+
+            ip_json_response = get_ip(machine_id) unless machine_data.key?(machine_id)
+            unless ip_json_response.nil?
+              machine_data[machine_id] = ["ip_address" => ip_json_response["lastIpAddress"],
+                                          "tags" => ip_json_response["machineTags"],
+                                          "status" => ip_json_response["healthStatus"]]
+            end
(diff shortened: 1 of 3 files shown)


tasks/connectors/ms_defender_tvm/ms_defender_tvm.rb
```diff
@@ -83,7 +83,6 @@ def self.metadata
 
       def run(opts)
```
mukhansk, line 84: It would be even better to split the run method into smaller methods. But it’s more about code style and readability.

tasks/connectors/ms_defender_tvm/ms_defender_tvm.rb
```diff
@@ -121,111 +120,120 @@ def run(opts)
...
-            end
-
-            if asset_id.!= machine_id
-              if asset_count == batch_page_size
-                submit_count += 1
-                print_debug "#{submit_count} about to upload file"
-                filename = "microsoft_tvm_kdi_#{submit_count}.json"
-                kdi_upload("#{$basedir}/#{output_directory}", filename, @kenna_connector_id, @kenna_api_host, @kenna_api_key, false, 3)
-                asset_count = 0
+              vuln_cve = vuln.fetch("cveId")
```
mukhansk, line 123: There are too many spaces (4). Rubocop is configured so that tab should be 2 spaces. When you fix it, the Lint Code Base check will pass

tasks/connectors/ms_defender_tvm/ms_defender_tvm.rb
```diff
@@ -121,111 +120,120 @@ def run(opts)
...
+                  "details" => JSON.pretty_generate(details)
+                }
+
+                # craft the vuln def hash
+                vuln_def = {
+                  "scanner_identifier" => scanner_id,
+                  "scanner_type" => "MS Defender TVM",
+                  "name" => vuln_name
+                }
+                vuln_def[:cve_identifiers] = vuln_cve.to_s if !vuln_cve.nil? && !vuln_cve.empty?
```
mukhansk, line 230: I would recommend using this instead
`vuln_def[:cve_identifiers] = vuln_cve.to_s unless vuln_cve.nil? && vuln_cve.empty?`

The result is the same, but we got rid of two objections.

tasks/connectors/ms_defender_tvm/ms_defender_tvm.rb
```diff
@@ -121,111 +120,120 @@ def run(opts)
...
+
+              machine_id = vuln.fetch("deviceId")
+              fqdn = vuln.fetch("deviceName")
+
+              ip_json_response = get_ip(machine_id) unless machine_data.has_key?(machine_id)
+
+              unless ip_json_response.nil?
+                machine_data[machine_id] = ["ip_address" => ip_json_response["lastIpAddress"], "tags" => ip_json_response["machineTags"], "status" => ip_json_response["healthStatus"]]
+              end
+              # Get the asset details & craft them into a hash
```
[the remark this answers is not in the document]
venkatawunnava: rubocop didn't like moving `unless`. I update the rest though

tasks/connectors/ms_defender_tvm/ms_defender_tvm.rb
```diff
@@ -121,111 +120,120 @@ def run(opts)
...
+              else
+                vuln_name = vuln_cve
+                vuln_cve = nil
+              end
+
+              machine_id = vuln.fetch("deviceId")
+              fqdn = vuln.fetch("deviceName")
+
+              ip_json_response = get_ip(machine_id) unless machine_data.has_key?(machine_id)
+
```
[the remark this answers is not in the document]
venkatawunnava: changed

tasks/connectors/ms_defender_tvm/ms_defender_tvm.rb
venkatawunnava: I didn't touch this code but let me know if would still be needed?

tasks/connectors/ms_defender_tvm/ms_defender_tvm.rb
venkatawunnava: done

tasks/connectors/ms_defender_tvm/ms_defender_tvm.rb
venkatawunnava: same this code too. planning to leave this as is

tasks/connectors/ms_defender_tvm/ms_defender_tvm.rb
```diff
@@ -133,9 +132,19 @@ def run(opts)
...
 
+            ip_json_response = get_ip(machine_id) unless machine_data.key?(machine_id)
+            unless ip_json_response.nil?
+              machine_data[machine_id] = ["ip_address" => ip_json_response["lastIpAddress"],
+                                          "tags" => ip_json_response["machineTags"],
+                                          "status" => ip_json_response["healthStatus"]]
+            end
             # Get the asset details & craft them into a hash
+
+            next unless machine_data[machine_id].first.dig("status") == "Active"
```
annapoltieva, line 143: Is there a reason why `["ip_address" => ip_json_response["lastIpAddress"], "tags" => ip_json_response["machineTags"], "status" => ip_json_response["healthStatus"]] `needs to be a hash inside an array and not just a hash? That way you could fetch a status like machine_data[machine_id]["status"]

annapoltieva, line 143: I think, this would error with `undefined method `first' for nil:NilClass (NoMethodError)` if `ip_json_response` is `nil`

annapoltieva (changes requested):

venkatawunnava:
Going to close this one. It was kind of implemented with pipeline and would very quickly hit the microsoft limit 

`Rate limitations for this API are 100 calls per minute and 1500 calls per hour.`

PR #419, opened 2022-08-01
Fix: guard to allow to execute without kenna api

fabiobcsouza:
That pr fixes a issue when trying to run a specific Veracode task without kenna api

Before:
Screen Shot 2022-08-01 at 13 00 09

Change:
Screen Shot 2022-08-01 at 13 00 37


mukhansk (approved):
Looks fine to me

PR #420, opened 2022-08-02
Fix illegal quoting for csv2kdi with liberal_parsing parameter

mukhansk:
Try to fix the error "illegal quoting.

Try №1
with `liberal_parsing: true`

Screenshot 2022-08-02 at 12 16 54


annapoltieva (approved):
Let's give it a try!

PR #423, opened 2022-08-11
Update edgescan connector

ShayVD:
# Context

The connector was not properly ingesting kenna assets created from IP address CIDR or block ranges.
A typo was also present where the "external_id" field was being passed in as "external_asset_id".

# Changes

Updated the split IP range functions to make use of IPAddr library.
Fixed "external_id" typo.

The updated connector has been tested successfully.


tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
```diff
@@ -28,8 +28,8 @@ def location_specifiers
         end
 
         def matching_location_specifier
-          location_specifiers.find { |specifier| specifier["id"] == location_specifier_id } ||
-            location_specifiers.find { |specifier| location.include?(specifier["location"]) }
+          location_specifiers.find { |specifier| not_ip_range?(specifier) && specifier["id"] == location_specifier_id } ||
```
whitfieldsdad, line 31: Why are we checking to see if an IP address is in an IP address range rather than directly comparing IP addresses? 

Isn't there already a 1:1 mapping between hosts and locations?

tasks/connectors/edgescan/lib/kenna_api.rb
```diff
@@ -15,9 +15,8 @@ def initialize(options)
 
         # Converts Edgescan location specifiers and vulnerabilities into Kenna assets and adds them to memory
         def add_assets(edgescan_location_specifiers, edgescan_vulnerabilities)
-          kenna_assets = edgescan_location_specifiers.map(&:to_kenna_asset).append(
-            edgescan_vulnerabilities.reject(&:matching_location_specifier).map(&:to_kenna_asset)
-          ).flatten
+          kenna_assets = edgescan_location_specifiers.map(&:to_kenna_asset).flatten.uniq!
+          kenna_assets.concat(edgescan_vulnerabilities.map(&:to_kenna_asset).uniq! - kenna_assets)
```
whitfieldsdad, line 19: Could we add some code comments here?

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
```diff
@@ -124,6 +124,10 @@ def url?
           !!(location =~ URI::DEFAULT_PARSER.make_regexp)
         end
 
+        def not_ip_range?(specifier)
```
whitfieldsdad, line 127: Why do we need to check if an IP address is in an IP address range?

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
ShayVD: Yes there is a 1:1 mapping as you say for hosts and vulnerabilities. This is more for vulnerabilities that do not have a corresponding host.

There is not always a 1:1 mapping between vulnerabilities and location specifier's locations. A locations specifier could have a URL location "https://test.example.com", and the vulnerability's location could be "https://test.example.com/some/further/page.html". Though the location is different we would want to group those vulnerabilities with that location specifier together.

If a host is unavailable, then the matching location specifier's location will be used as the vulnerability's external ID, and I did not want an external ID with an IP address range in it.

For example:
Host is available: external_id => "ES1234 {host.location}"
Host is unavailable, but there is a matching location specifier => "ES1234 {location_specifier.location" unless the location is an IP address range
Host and location specifier are unavailable => "ES1234 {vulnerability.location}"

tasks/connectors/edgescan/lib/edgescan_vulnerability.rb
ShayVD: We're checking that a vulnerability's location specifier is not an ip address range so that we don't use it in the vulnerability's external_id. See comment above.

tasks/connectors/edgescan/lib/kenna_api.rb
ShayVD: Fixed in [80d3673](https://github.com/KennaSecurity/toolkit/pull/423/commits/80d367334c28868936fff425357249bafd6dadd2)

whitfieldsdad (dismissed):

ShayVD:
Hey @mukhansk can we get a review on this PR please.

tasks/connectors/edgescan/lib/kenna_api.rb
```diff
@@ -11,13 +11,18 @@ def initialize(options)
           @kenna_api_key = options[:kenna_api_key]
           @kenna_connector_id = options[:kenna_connector_id]
           @output_dir = "#{$basedir}/#{options[:output_directory]}"
+          @skip_autoclose = false
+          @max_retries = 3
+          @kdi_version = 1
         end
```
mukhansk, line 17: Hi) I have reviewed the changes in this connector and all these changes look reasonable. There is one point that I am currently trying to investigate - the possibility of using the KENNA KDI format version 2. So I think it would be better to use version 2

ShayVD: Yeah I don't see why not. Tested and updated in 3f92116

mukhansk (approved):
You did a great job!

PR #424, opened 2022-08-11
retry 422, treat success field as string

sjbridge:
## SUP-969

# Problem
On the connector file upload endpoint Kenna returns a 422 in a very generic failure scenario when the file could not be persisted in Kenna. This can be a sporadic problem, so we should not immediately fail toolkit run attempts when this happens..

Another failure saw errors when we try to process a seeming empty response from Kenna. The current response processing doesn't correctly check for success in the response so some failures slip by

# Solution
-add UnprocessableEntity to a rescue statement with retries.
-check "success" field assuming it's string value
-treat lack of "success":"true" in response as a generic failure (which does retries)

### DIFF
--- a/lib/api/client.rb
+++ b/lib/api/client.rb
@@ -152,12 +152,11 @@ module Kenna
           )
 
           query_response_json = JSON.parse(query_response.body)
-          if query_response_json.fetch("success")=="true"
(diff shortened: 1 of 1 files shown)


annapoltieva (approved):
LGTM

PR #425, opened 2022-08-18
SUP-980-sentinel_fix_params_and_remove_api_check_step

sjbridge:
# SUP-980

## Problem
Client was seeing 401 unauthorized errors using the Whitehat/Sentinel toolkit task despite providing valid api token. Support discovered that the params provided to RestClient were not in the right form. After resolving that I ran into two other problems:
1. The task tries to "validate" the api token using an endpoint that doesn't seem to be documented and was returning 404
2. Accept header was not passed in correctly, so we were getting data back in unexpected XML

## Solution
1. fix params format to be {header1: value1, header2: value2, params:{urlparam1: "value1", urlparam2: "value2"}}
2. remove invalid validation call step - we can expect a 401 in later calls if there's a problem with the token
  - see: `https://apidocs.whitehatsec.com/whs/reference/`
  - was using a GET to '/' endpoint
3. specify accept: header explicitly

### DIFF
--- a/tasks/connectors/ntt_sentinel_dynamic/lib/api_client.rb
+++ b/tasks/connectors/ntt_sentinel_dynamic/lib/api_client.rb
@@ -17,13 +17,6 @@ module Kenna
           @page_size = page_size
         end
 
-        def api_key_valid?
-          get("/", retries: 0)
-          true
-        rescue Error
-          false
-        end
-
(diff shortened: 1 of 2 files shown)


nhenry012 (approved):
If the key is invalid the actual api call should return a 401. I do not think the additional validation for the token was needed.

PR #426, opened 2022-08-24
Update readme.md

caleb-eckenwiler:
Tags don't exist in findings. The item listed implies they are. That is incorrect. Adjusting the readme to be more accurate, moving Scan_type to additional information mapping along with adding additional data points that are mapped into addtl info.


mukhansk (approved):
This is a good remark

PR #434, opened 2022-09-22
Successful run snyk with new token

mukhansk:
Jira: [CON-1767](https://kennasecurity.atlassian.net/browse/CON-1767)

Fixes bugs that arise in the Lewis's clients.

Screenshots:
Screenshot 2022-09-22 at 12 49 43
Screenshot 2022-09-22 at 12 50 19
Screenshot 2022-09-22 at 17 48 40

### DIFF
--- a/tasks/connectors/snyk_v2/snyk_v2_task.rb
+++ b/tasks/connectors/snyk_v2/snyk_v2_task.rb
@@ -111,7 +111,6 @@ module Kenna
           types << "license" if @include_license
 
           while more_pages
-            page_num += 1
             issue_json = []
 
             project_ids.each_slice(500) do |sliced_ids|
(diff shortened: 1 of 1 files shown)


tasks/connectors/snyk_v2/snyk_v2_task.rb
```diff
@@ -125,9 +124,11 @@ def run(opts)
               }"
               print_debug "issue filter json = #{issue_filter_json}"
 
-              issue_json << client.snyk_get_issues(@page_size, issue_filter_json, page_num, @from_date, @to_date)
-              print_debug "issue json = #{issue_json}"
+              page_num += 1
+              issue_json << client.snyk_get_issues(@page_size, issue_filter_json, page_num, @from_date, @to_date) unless
```
ricardomontesCsc, line 128: The code looks good, but you could reduce the large of this line doing something like this:

`issue_json << client.snyk_get_issues(@page_size, issue_filter_json, page_num, @from_date, @to_date).compact`

this is gonna remove all `nil` values what would come from the method, so at the end of the code block the `array` will be flattened so that´s gonna remove all empty values.

ricardomontesCsc (approved):

tasks/connectors/snyk_v2/snyk_v2_task.rb
mukhansk: Thanks, that's a good idea, but it doesn't work in this case

PR #435, opened 2022-10-06
account for nil tags with the Wiz toolkit

tbowyer:
Summary

Kenna Support received a ticket about the Wiz toolkit failing due to this error
`/opt/app/toolkit/tasks/connectors/wiz/lib/vulns_mapper.rb:34:in `extract_tags': undefined method `map' for nil:NilClass (NoMethodError)`

After adding debug statements I found that some WIz vulnerableAsset objects have no tag info returned via the API.  I met with Wiz/SYSCO to discuss this and Wiz confirmed that in some rare cases, the payloads will have `nil` for the vulnerableAsset["tags"] field.  This change will force the toolkit to check for the existence of tags before it tries to extract_tags.

I tested the changes against my sandbox environment and valid KDI files were produced. After uploading the files they created valid asset/vuln records.


sjbridge (approved):
Looks like a fairly safe change to me.

venkatawunnava:
closing this pr and created a new one https://github.com/KennaSecurity/toolkit/pull/436

PR #450, opened 2022-12-20
better CVE regex

sjbridge:
# Problem
Wiz tookit connnector runs are sometimes failing due to invalid CVE identifiers: https://admin.sandbox.us.kennasecurity.com/admin/connector_runs/1621311

From looking at the failed data I found that Wiz provides definitions like "CVE-2017-NONE". The current regex we use matches the first part of this identifier.

# Solution
Improve the CVE matching regex we use so it only gets true CVEs


DegzKS (dismissed):
LGTM

BlakeHendersonKS (approved):

BlakeHendersonKS (approved):

PR #463, opened 2023-03-01
Get EC2 instance ID in WIZ

nhenry012:
Gets the EC2 Id in the Wiz connector if it is available and the cloud platform is AWS


algirald (approved):

sjbridge (approved):
LGTM

PR #464, opened 2023-03-06
Lacework rate limits & retry custom wait

sjbridge:
# Problem
Lacework has API rate limiting at the rate of 480 requests per hour. Clients with enough data will hit this and currently the task doesn't have logic to work around this limitation, so the toolkit task will error out.

# Solution
Added the limiter gem to the toolkit and applied a rate of 480 per hour to get/post methods that retrieve data. Set to spread out requests.

Lacework makes use of a response header "".  This is a draft standard proposal from IETF, so I felt comfortable putting it's use in a shared http utility library of the toolkit provided it's not made required.

# Testing performed

### DIFF
--- a/tasks/connectors/lacework/lacework.rb
+++ b/tasks/connectors/lacework/lacework.rb
@@ -64,6 +64,7 @@ module Kenna
       end
 
       def run(opts)
+        RestClient.log=STDOUT
         super
 
         lacework_account = @options[:lacework_account]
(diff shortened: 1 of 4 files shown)


tasks/connectors/lacework/lacework.rb
```diff
@@ -71,6 +74,8 @@ def run(opts)
         @kenna_connector_id = @options[:kenna_connector_id]
         @kenna_api_host = @options[:kenna_api_host]
 
+        @ratequeue = Limiter::RateQueue.new(680, interval: 3600, balanced: true)
```
rick-kenna, line 77: It would be nice to have a comment here on what the Limiter values are for.  680?  3600?

tasks/connectors/lacework/lib/lacework_helper.rb
```diff
@@ -36,27 +37,16 @@ def generate_temporary_lacework_api_token(account, api_key, api_secret)
       end
 
       def lacework_post(url:, body:, api_token:)
-        uri = URI.parse(url)
+        ratequeue.shift
+        headers = { 'Authorization' => "Bearer #{api_token}", 'Content-type' => 'application/json' }
+        response = http_post(url, headers, body, 3)
 
-        request = Net::HTTP::Post.new(uri)
```
rick-kenna, line 41: Good replacement.

sjbridge:
> Where is `RateLimit-Reset` set in the header?

This is a response header provided by Lacework. It's currently a draft standard by IETF: https://datatracker.ietf.org/doc/draft-ietf-httpapi-ratelimit-headers/

rick-kenna (approved):

PR #467, opened 2023-03-10
Remove json-write-stream gem

jgarber-cisco:
The gem `json-write-stream` depends on [`json_pure@1.8.6`](https://github.com/flori/json/tree/7f4cfd853f2c919d854fb95548a19980feff17e8)  (2017... "Try to be compatible with ruby 1.8"), raising an error:
```
ArgumentError:
  wrong number of arguments (given 2, expected 1)
# /usr/local/bundle/gems/json_pure-1.8.6/lib/json/common.rb:155:in `initialize'
# /usr/local/bundle/gems/json_pure-1.8.6/lib/json/common.rb:155:in `new'
# /usr/local/bundle/gems/json_pure-1.8.6/lib/json/common.rb:155:in `parse'
...
# /usr/local/bundle/gems/jmespath-1.6.2/lib/jmespath.rb:6:in `<top (required)>'
# /usr/local/bundle/gems/aws-sdk-core-3.168.4/lib/aws-sdk-core.rb:5:in `require'
...
# ./tasks/connectors/aws_guardduty/aws_guardduty.rb:3:in `<top (required)>'
# ./lib/toolkit.rb:46:in `require_relative'
# ./lib/toolkit.rb:46:in `block in <top (required)>'
# ./lib/toolkit.rb:46:in `each'
# ./lib/toolkit.rb:46:in `<top (required)>'
```

This gem appears unnecessary to me because:
* We're writing to a file, not to a stream.
* Uploading to Kenna doesn't happen until the entire file is written.
* Large datasets are handled by processing in batches.

I've written a spec for `write_file_stream` to prevent regression, but with so little test coverage, I can't guarantee it won't break something.


nhenry012 (approved):
This seems significantly simplified.

PR #473, opened 2023-03-31
[CON-2806] Add detailedName and detectionMethod to the GraphQL query.

jgarber-cisco:
In #470 we added mapping for detailedName and detectionMethod without adding those fields to the query. This fixes that.


ricardomontesCsc (approved):
LGFM 👍🏼

PR #477, opened 2023-04-07
Update readme.md

lidiaattalla:
Changing lacework_api_token to lacework_api_key
Clarifying the account option, it just needs to be the subdomain of their api URL (account instead of account.lacework.net for example)


lidiaattalla:
Based on Trevor's tests

jaredkalmus (approved):

PR #480, opened 2023-04-14
add omit_line_number option

sjbridge:
### Problem

A client has reported that a number of their Kenna findings not closing after subsequent task runs after being resolved in Veracode. The reason seems to be that the findings are just dropping off of the api without ever showing a closed status in the data.  Since the file locators with line number are usually specific to their particular findings, this means we never see the asset again if the findings falls of the api. If we never see the asset again there's no chances for auto-close functionality to take effect.

### Solution

Provide an optional (to not break anyone's workflow) task option that when set causes us to omit the line number from file locator on non-SCA findings.

### DIFF
--- a/tasks/connectors/veracode_findings/veracode_findings.rb
+++ b/tasks/connectors/veracode_findings/veracode_findings.rb
@@ -53,7 +53,7 @@ module Kenna
               type: "boolean",
               required: false,
               default: false,
-              description: "When set true, it will omit the line number at the end of file_locator."
+              description: "When set to true, it will omit the line number at the end of file_locator."
             }
           ]
         }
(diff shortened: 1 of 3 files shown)


tasks/connectors/veracode_findings/veracode_findings.rb
```diff
@@ -47,8 +47,14 @@ def self.metadata
...
               default: "output/veracode",
-              description: "If set, will write a file upon completion. Path is relative to #{$basedir}" }
-
+              description: "If set, will write a file upon completion. Path is relative to #{$basedir}" },
+            {
+              name: "omit_line_number",
+              type: "boolean",
+              required: false,
+              default: false,
+              description: "When set true, it will omit the line number at the end of file_locator."
```
jgarber-cisco, line 56: "When set **to** true" or "When true"

jgarber-cisco (changes requested):

jgarber-cisco (approved):

PR #481, opened 2023-04-18
Veracode findings put omit lines in scope

sjbridge:
## Problem
Recently implemented omit_line_number option is not in scope where used.

## Solution
At the beginning of the task run pull it out of the options hash and pass it as an argument into the method where it is used.


BlakeHendersonKS (dismissed):
Including the param will help, apologies I didn't catch the last one.

BlakeHendersonKS (approved):
omit_line_number is omitting better!

PR #484, opened 2023-04-25
SUP-1171-lacework_map_vuln_description

sjbridge:
## Problem
With the set of fields the Lacework toolkit task brings in, vulns are scored, but technically informational, and don't have a visible description in the UI

## Solution
Map in 'description' vuln definition field. Map in 'cve_identifiers' vuln definition field.

image

- cve description visible in Explore

image

- vuln description visible on vuln detail page


BlakeHendersonKS (approved):
Looks like a reasonable and small change.

PR #485, opened 2023-04-28
Con 2646/3024 snyk loop vuln names

nhenry012:
Loop over each possible vuln name and create a new finding. This will allow those findings to be searchable individually in appsec.


jgarber-cisco (approved):

DegzKS (dismissed):
LGTM

PR #487, opened 2023-05-04
Con 2926 veracode unique scanner ids

nhenry012:
CON-2926

update the scanner_id for fixes to use the app name and the unique issue ID. This does not unfortunately change SCA findings because those do not appear to have the required issue_id. However, those are not the findings causing our customer support ticket that leaded to this ticket. If we need to address that in the future we will do so in a new ticket.

### DIFF
--- a/spec/rspec_helper.rb
+++ b/spec/rspec_helper.rb
@@ -8,6 +8,7 @@ require_relative "../lib/toolkit"
 
 require "timecop"
 require 'vcr'
+require 'webmock/rspec'
 
 VCR.configure do |config|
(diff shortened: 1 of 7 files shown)


jgarber-cisco (dismissed):

DegzKS (approved):
LGTM.

NOTE: I think Webmock better fit to this project than VCR. Nice work!

algirald (dismissed):

ricardomontesCsc (dismissed):

algirald (approved):

PR #489, opened 2023-06-30
Toolkit readme update

slubeckatkenna:
README update with usage note regarding new(ish) base64 API keys that contain "$" characters will require escaping when called via shell

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ In this case, the expanse task:
 
     docker pull kennasecurity/toolkit && docker run -it kennasecurity/toolkit task=kenna_api_key_check kenna_api_key=$KENNA_API_KEY
 
-NOTE: Assuming a LINUX/UNIX shell is invoking the above, any '$' characters contained in API key must be escaped, e.g.: `KENNA_API_KEY="123456ABCDEFG\$"`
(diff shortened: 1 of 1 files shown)


rick-kenna (changes requested):
Agreed, but this also works: `export='123456ABCDEF$'`.  I think this is simplier and should be used instead or at least both should be mentioned.  You can also point to the [API Authentication](https://apidocs.kennasecurity.com/reference/api-authentication) section.

jgarber-cisco (dismissed):
I agree with @rick-kenna that it's preferable to encourage customers to single-quote their API keys in a shell rather than escaping them. It's much more fool-proof!

rick-kenna (approved):

rick-kenna:
@jgarber-cisco How is it cryptic?  It is a reference.  And the API documentation discusses the exact point with an example for Linux and Windows.

jgarber-cisco:
Ok, sorry to be over the top!
________________________________

slubeckatkenna:
Ideally we should probably point to Rick's https://github.com/rick-kenna/API_keys_safe_store over storing/exporting plaintext keys but linking out to apidocs should suffice for now. @jgarber-cisco  May I dismiss your changes review in order to merge?

jgarber-cisco (approved):

PR #495, opened 2023-07-27
Add ability to specify time as delta in seconds for Cylera

bodubas:
1. Added ability to specify relative time in seconds for time parameters, e.g. cylera_last_seen_after=60s
2. Reversed VCR back as tests are failing without it.

### DIFF
--- a/spec/tasks/connectors/cylera/cylera_task_spec.rb
+++ b/spec/tasks/connectors/cylera/cylera_task_spec.rb
@@ -50,5 +50,24 @@ RSpec.describe Kenna::Toolkit::CyleraTask do
         expect { task.run(options) }.to raise_error(SystemExit) { |e| expect(e.status).to_not be_zero }
       end
     end
+
+    context 'when time value is unexpected' do
+      let(:options) do
+        {
+          cylera_api_host: 'cylera.host',
+          cylera_api_user: 'api_user',
+          cylera_api_password: 'api_pass',
+          kenna_api_key: 'api_key',
+          kenna_api_host: 'kenna.example.com',
+          kenna_connector_id: '12',
+          cylera_last_seen_after: '60m'
+        }
+      end
+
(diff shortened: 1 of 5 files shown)


tasks/connectors/cylera/lib/client.rb
jgarber-cisco, line 1: Why is VCR required in the client?

spec/tasks/connectors/cylera/cylera_task_spec.rb
```diff
@@ -50,5 +50,23 @@
         expect { task.run(options) }.to raise_error(SystemExit) { |e| expect(e.status).to_not be_zero }
       end
     end
+
+    context 'when time value is unexpected' do
```
jgarber-cisco, line 54: "when time value is an unexpected format"?

tasks/connectors/cylera/lib/vcr.rb
jgarber-cisco, line 1: I think I deleted this file once before and everything was fine. 😁  VCR should only be used in specs, right?

jgarber-cisco (changes requested):

spec/tasks/connectors/cylera/cylera_task_spec.rb
bodubas: Updated

jgarber-cisco (approved):
Can we remove the VCR thing next week? I really don't like it being in lib. 
When I removed it before, I thought the tests passed because it was in `spec/rspec_helper.rb`. Let's work on it together if not.

bodubas:
> Can we remove the VCR thing next week? I really don't like it being in lib. When I removed it before, I thought the tests passed because it was in `spec/rspec_helper.rb`. Let's work on it together if not.

Sure, thanks for the review. Will work on this next week.

PR #497, opened 2023-08-01
Incremental run for cylera connector

Snick555:


tasks/connectors/cylera/cylera_task.rb
```diff
@@ -204,6 +211,12 @@ def run(opts)
 
         initialize_options
 
+        if @incremental
+          api_client = Kenna::Api::Client.new(@kenna_api_key, @kenna_api_host)
+          connector_runs = api_client.get_connector_runs(@kenna_connector_id)[:results]
+          @inventory_devices_params[:last_seen_after] = connector_runs.find { |e| e['success'] }.try(:[], 'start_time')&.to_datetime.to_i
```
bodubas, line 217: I think we need to make this less precise to avoid missing data. I propose to subtrack 1 day from the last_connector_run for now. The reason for this is that toolkit kicks off Kenna connector only when all the assets are pulled instead of the toolkit start time. So it might have large delay until toolkit pulls all data from Cylera.

But, if we declaire that incremental option works only for tasks that run less then 24 hours, we can subtract 24 hours from the last successful run returned from Kenna API. As a result we will pull data for up to 2 days, but won't loose any as far as pulling data from Cylera won't take more then 24 hours.

spec/tasks/connectors/cylera/cylera_task_spec.rb
```diff
@@ -17,14 +17,21 @@
...
+      )
     end
 
+    before { allow(Kenna::Api::Client).to receive(:new) { kenna_client } }
+
     it 'succeeds' do
-      expect { task.run(options) }.to_not raise_error
+      VCR.use_cassette('cylera') do
+        expect { task.run(options) }.to_not raise_error
+      end
```
jgarber-cisco, line 34: You might be able to do this once and not have to `use_cassette` every time?

```
around(:example) do |ex|
  VCR.use_cassette('cylera') do
    ex.run
  end
end
```

jgarber-cisco (approved):
Nice! I left one idea for drying up `VCR.use_cassette` but just a suggestion you can take or leave.

PR #499, opened 2023-08-09
Add tags and locators for Cylera

bodubas:
It was discovered that there are more fields that can be mapped to Kenna. Added those fields in this PR.


jgarber-cisco (commented):
Should there be spec changes as well?

jgarber-cisco (approved):

PR #510, opened 2023-10-13
Update readme AWSV2 Skip_autoclose and Ignore_scanner_last_seen_time info

FPolillo:
This is an updated requested by Yi Chen Cheng in ticket TD-302.


jgarber-cisco (changes requested):
These are minor details that should not be at the top of the README. They're more in line with the limitations at the bottom.

The header should tell the reader what the notes below it are, not just repeat their identification. The notes on those settings seem like hacks or "gotchas," but I'm sure there's a more professional way to say it, which I leave to the content professionals.

PR #515, opened 2023-11-28
Sup 1399 code scanning error

effany:
Created a new toolkit task called "github_code_scanning_for_org" in order to separate with "github_code_scanning_task" This is due to these 2 tasks are using different endpoint and it's better to separate them to avoid confusion when passing in the options from customer's end. For "github_code_scanning_for_org" you only need to pass in the organization name option. For "github_code_scanning_task" you'll need to pass in Owner/Repositories option. 

We can refer the below 2 articles which state the different endpoints
[(https://docs.github.com/en/rest/code-scanning/code-scanning?apiVersion=2022-11-28#list-code-scanning-alerts-for-a-repository])

[(https://docs.github.com/en/free-pro-team@latest/rest/code-scanning/code-scanning?apiVersion=2022-11-28#list-code-scanning-alerts-for-an-organization )]

The running options shall be something like this: 
`docker run -it --rm -v /PATH/TO/LOCAL/DIR:/PATH/IN/DOCKER toolkit:latest task=github_code_scanning_for_org github_username=USERNAME github_token=github_pat_REDACT github_organization=ORGNAME`


effany:
refactor the code instead of create new task 
https://github.com/KennaSecurity/toolkit/pull/517

PR #516, opened 2023-11-28
update the task so that when there's no path, it input no path found …

effany:
Some of the data sent back from github secret scanning is missing the "path" information. Therefore, adding a clause that if there's no path, input default as "no path found"


sjbridge (approved):
LGTM

PR #517, opened 2023-11-29
Sup 1399 code scanning error refactor code

effany:
Modify the toolkit task called "github_code_scanning" in order to accommodate the options for organisations. 

This is due to Owner/repositories and Organisation names are using different endpoint. 
and we need to accomodate both options pass in by the customers. 

To scan the repositories from an org, you only need to pass in the organisation name option. For scanning the owner's repositories,  you'll need to pass in Owner/Repositories option.

We can refer the below 2 articles which state the different endpoints
[(https://docs.github.com/en/rest/code-scanning/code-scanning?apiVersion=2022-11-28#list-code-scanning-alerts-for-a-repository])

[(https://docs.github.com/en/free-pro-team@latest/rest/code-scanning/code-scanning?apiVersion=2022-11-28#list-code-scanning-alerts-for-an-organization )]

The running options shall be something like this:
docker run -it --rm -v /PATH/TO/LOCAL/DIR:/PATH/IN/DOCKER toolkit:latest task=github_code_scanning_for_org github_username=USERNAME github_token=github_pat_REDACT github_organization=ORGNAME

or 

docker run -it --rm -v /PATH/TO/LOCAL/DIR:/PATH/IN/DOCKER toolkit:latest task=github_code_scanning_for_org github_username=USERNAME github_token=github_pat_REDACT github_repositories=Owner/Repo


tasks/connectors/github_code_scanning/readme.md
```diff
@@ -6,12 +6,15 @@ To run this task you need the following information from GitHub:
...
 2. GitHub token 
-3. One or more repository names
+3. One or more repository/organization names
+
 
 **IMPORTANT: you must be an administrator for the repository or organization, and you must use an access token with the repo scope or security_events scope.
 If GitHub 2FA is enabled, the access token MUST be configured for SSO. 
 If the user is not an admin at the time of Key generation, you will need to promote the user to admin, and regenerate the key. **
 
+**IMPORTANT: you can only input your organizations names or repositories list in one command. You CANNOT provide both organizations and repositories at the same time. Providing the organization names will return all the repositories scan from that organization. Providing the repositories will return the scans from the selected repositories from that individual owner. You CANNOT provide Organisation Name + Repositories. Ex. KennaSecurityOrg/repo1. You can only provide Owner + Repositories. Ex. KennaSecurity/repo1. **
```
BlakeHendersonKS, line 16: Thank you.

tasks/connectors/github_code_scanning/github_code_scanning_task.rb
```diff
@@ -82,9 +87,18 @@ def run(opts)
           initialize_options
           initialize_client
 
-          @repositories.each do |repo|
-            endpoint = "/repos/#{repo}/code-scanning/alerts"
-            import_alerts(repo, endpoint)
+          if !@repositories.empty? && @organizations.empty?
```
BlakeHendersonKS, line 90: I don't know why we don't just scan both if either are present, but this should work I suppose.

BlakeHendersonKS (approved):
We could modify it to read from both simultaneously, but this looks like it would work.

sjbridge (approved):

sjbridge (approved):

PR #521, opened 2023-12-12
SUP-1426 Correct default page size for wiz issues

dianaroz:
## Summary

**Jira**: [SUP-1426](https://kennasecurity.atlassian.net/browse/SUP-1426)

### Problem:
Running wiz task with import_type ALL or ISSUES with default wiz_page_size (5000) results in error "'first' must be less than or equal to 500" during issues import. As per wiz documentation, maximum "first" value for issues is 500 while for vulnerabilities it's 5000.

### Solution:
Implement separate parameters for issues page size and vulnerabilities page size.

[SUP-1426]: https://kennasecurity.atlassian.net/browse/SUP-1426?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tasks/connectors/wiz/lib/wiz_client.rb
+++ b/tasks/connectors/wiz/lib/wiz_client.rb
@@ -6,9 +6,10 @@ module Kenna
       class Client
         class ApiError < StandardError; end
 
-        def initialize(client_id, client_secret, auth_endpoint, api_host, page_size, days_back, vuln_object_types, severity, issue_status)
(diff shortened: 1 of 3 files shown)


sjbridge (approved):
LGTM

jgarber-cisco (commented):
I like the changes to the code, but how disruptive will the CLI change be for users? Should we keep the wiz_page_size parameter available and have it populate vuln_page_size and issue_page_size if those more specific options aren't provided? I hate to see us break things for customers that aren't experiencing this problem. (Disregard if there's just one or two customers using it.)

PR #524, opened 2023-12-18
Github Dependabot Improvements

shotop:
**Primary Problem**:  The current implementation of the Github Dependabot Toolkit connector yields asset and vulnerability data within Kenna that doesn't align conceptually with the data that exists in Github.  In Github there are repositories under which exist many vulnerabilities.  In Kenna, this structure would align with an asset that has many vulnerabilities.  Currently, when the Toolkit connector is run, it creates a new asset in Kenna for every single vulnerability coming from Github:

Screenshot 2023-12-18 at 2 41 47 PM

**Solution**: What we should do is create a single Asset in Kenna for each Repository in Github and under each of those assets, there should be many vulnerabilities.  A one-to-many relationship, vs a one-to-one relationship. This way, each repo can be viewed as a single asset and its risk can be understood and mitigated in a way that is similar to all other assets in Kenna. 

Screenshot 2023-12-18 at 2 53 35 PM

**Additional Issues**:The current implementation also brings in all vulnerabilities, including closed vulnerabilities, and marks them as Open.  This PR limits the import to true Open vulnerabilities only.  Important to note that this update will cause what appear to fewer vulns brought into Kenna, but in reality it is bringing in only truly Open vulnerabilities.  Confirmed this during QA with @venkatawunnava . 

**Confirming All Repo Data is Imported:** 

Screenshot 2024-01-08 at 11 51 55 AM

Screenshot 2024-01-08 at 11 52 32 AM


cyberhiker (approved):

sjbridge (approved):
LGTM

venkatawunnava (approved):
tested the changes in https://kennasecurity.atlassian.net/browse/CON-4508
Looks good

PR #528, opened 2024-01-19
refactor the code with adding options

effany:
The order logic here indicate that we're selecting the svd based on the match between scanner identifier and external_unique_id.

https://github.com/KennaSecurity/beehive/blob/master/lib/beehive/sharded/finding.rb#L85C25-L85C25

    @svd = q.where('scanner_vulnerability_definitions.created_at <= ?', created_at)
            .order([
                     'scanner_vulnerability_definitions.identifier = ? DESC',
                     [external_unique_id]
                   ])
            .first
Based on the original mapping in the toolkit, we're providing different vuln_def, vul_def_name then the scanner_identifier which are the 2 elements composing the svd identifier and external_unique_id. This sometimes cause the mismatch of the svd to findings.

Proposing to provide an options in the toolkit to  make the vuln_def / vuln_def_name in the finding hash the same as the scanner_identifier to match the selection logic in our importer code.

### DIFF
--- a/tasks/connectors/qualys_was/README.md
+++ b/tasks/connectors/qualys_was/README.md
@@ -46,7 +46,7 @@ Run the QualysWas task following the guidelines on the main [toolkit help page](
 | kenna_api_host | hostname | false | Kenna API Hostname |
 | kenna_connector_id | integer | false | If set, we'll try to upload to this connector |
 | output_directory | filename | false | Will alter default filename for output. Path is relative to #{$basedir} |
-| match_ScannerIdentifier_with_vuln_def_name | boolean | false | If true, the finding scanner_identifier and the vuln_def name will match the format 'QID - ID - Name'. If a connector first ran without this parameter and it is enabled in a later run, it will result in automatic resolution of the existing findings with the previous format ('QID - ID') and creation of new findings with the new format within CVM. |
+| match_finding_with_vuln_def | boolean | false | If true, the finding scanner_identifier and the vuln_def name will match the format 'QID - ID - Name'. If a connector first ran without this parameter and it is enabled in a later run, it will result in automatic resolution of the existing findings with the previous format ('QID - ID') and creation of new findings with the new format within CVM. |
 
 
 ## Example Command Line:
--- a/tasks/connectors/qualys_was/README.md
+++ b/tasks/connectors/qualys_was/README.md
@@ -46,7 +46,7 @@ Run the QualysWas task following the guidelines on the main [toolkit help page](
 | kenna_api_host | hostname | false | Kenna API Hostname |
 | kenna_connector_id | integer | false | If set, we'll try to upload to this connector |
 | output_directory | filename | false | Will alter default filename for output. Path is relative to #{$basedir} |
-| match_ScannerIdentifier_with_vuln_def_name | boolean | false | If true, the Scanner Description in UI will be coming from the vuln_defs' name in original JSON file matched with scanner_identifier. Each imported finding unless it has the exact same qid, id, and name from source data, will be created as new one. |
+| match_ScannerIdentifier_with_vuln_def_name | boolean | false | If true, the finding scanner_identifier and the vuln_def name will match the format 'QID - ID - Name'. If a connector first ran without this parameter and it is enabled in a later run, it will result in automatic resolution of the existing findings with the previous format ('QID - ID') and creation of new findings with the new format within CVM. |
 
 
 ## Example Command Line:
--- a/tasks/connectors/qualys_was/README.md
+++ b/tasks/connectors/qualys_was/README.md
@@ -46,9 +46,10 @@ Run the QualysWas task following the guidelines on the main [toolkit help page](
 | kenna_api_host | hostname | false | Kenna API Hostname |
 | kenna_connector_id | integer | false | If set, we'll try to upload to this connector |
 | output_directory | filename | false | Will alter default filename for output. Path is relative to #{$basedir} |
+| match_ScannerIdentifier_with_vuln_def_name | boolean | false | If true, the Scanner Description in UI will be coming from the vuln_defs' name in original JSON file matched with scanner_identifier. Each imported finding unless it has the exact same qid, id, and name from source data, will be created as new one. |
 
 
 ## Example Command Line:
 
     toolkit:latest task=qualys_was qualys_was_domain=qualysapi.qg3.apps.qualys.com qualys_was_user=xxx qualys_was_password=xxx
-    qualys_was_api_version_url=/qps/rest/3.0/ qualys_was_score_filter=2 kenna_connector_id=15xxxx kenna_api_host=api.kennasecurity.com kenna_api_key=xxx
+    qualys_was_api_version_url=/qps/rest/3.0/ qualys_was_score_filter=2 kenna_connector_id=15xxxx kenna_api_host=api.kennasecurity.com kenna_api_key=xxx match_ScannerIdentifier_with_vuln_def_name=true
--- a/tasks/connectors/qualys_was/README.md
+++ b/tasks/connectors/qualys_was/README.md
@@ -52,4 +52,4 @@ Run the QualysWas task following the guidelines on the main [toolkit help page](
 ## Example Command Line:
 
     toolkit:latest task=qualys_was qualys_was_domain=qualysapi.qg3.apps.qualys.com qualys_was_user=xxx qualys_was_password=xxx
-    qualys_was_api_version_url=/qps/rest/3.0/ qualys_was_score_filter=2 kenna_connector_id=15xxxx kenna_api_host=api.kennasecurity.com kenna_api_key=xxx match_ScannerIdentifier_with_vuln_def_name=true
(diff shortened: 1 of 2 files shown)


tasks/connectors/qualys_was/qualys_was.rb
```diff
@@ -81,7 +81,12 @@ def self.metadata
               type: "filename",
               required: false,
               default: "output/qualys_was",
-              description: "If set, will write a file upon completion. Path is relative to #{$basedir}" }
+              description: "If set, will write a file upon completion. Path is relative to #{$basedir}" },
+            { name: "match_ScannerIdentifier_with_vuln_def_name",
```
dianaroz, line 85: Camel case is not used in toolkit parameters, for consistency I would suggest to use `match_scanner_identifier_with_vuln_def_name`

Alternatively, if we want to shorten it we could use something like `match_scan_id_with_vuln_def_name` or something a bit more general like `match_finding_with_vuln_def ` which speaks to the effect rather than the code changes.

tasks/connectors/qualys_was/qualys_was.rb
```diff
@@ -81,7 +81,12 @@ def self.metadata
...
+              type: "boolean",
+              required: false,
+              default: false,
+              description: "If true, the Scanner Description in UI will be coming from the vuln_defs' name in original JSON file matched with scanner_identifier. Each imported finding unless have the exact same qid, id, and name from source data, will be created as new one." }
```
dianaroz, line 89: This looks like it could be confusing for customers as it's describing a backend process.
I'd suggest something like this but interested in other people's feedback:

"If true, the finding scanner_identifier and the vuln_def name will match the format 'QID - ID - Name'. If a connector first ran without this parameter and it is enabled in a later run, it will result in automatic resolution of the existing findings with the previous format ('QID - ID') and creation of new findings with the new format within CVM."

tasks/connectors/qualys_was/qualys_was.rb
```diff
@@ -230,7 +235,14 @@ def domain_detail(find_from)
       end
 
       def name(find_from)
-        "#{find_from['qid']}-#{find_from['name']}"
+        if @options[:match_ScannerIdentifier_with_vuln_def_name]
+          return if find_from.nil?
```
dianaroz, line 239: curious about what would result in find_from being nil

I would say we'd have to return something either way, ingesting nil as the vuln_def name/finding scanner_identifier is likely to cause issues in CVM

tasks/connectors/qualys_was/qualys_was.rb
```diff
@@ -230,7 +235,14 @@ def domain_detail(find_from)
...
+
+          structured_name = [find_from['qid'], find_from['id'], find_from['name']].compact.join('-')
+          structured_name unless structured_name.empty?
```
dianaroz, line 242: I'm not sure this is necessary, I'd expect at least finding ID to be present for all findings
If QID/ID/Name were simultaneously resulting in nil values it could indicate a larger issue and we might want to catch that failure, if not, we'd still want to return something in here

tasks/connectors/qualys_was/qualys_was.rb
effany: thanks! I will update the code according to the suggestion :)

tasks/connectors/qualys_was/qualys_was.rb
effany: Thank you ! Your concern makes sense.

tasks/connectors/qualys_was/qualys_was.rb
effany: So in the code, find_from was defined as 

`find_from = data["Finding"]`
and 
data is from 
`finding["data"]`
finding is from findings which a collection from 
`findings_response = qualys_was_get_webapp_findings(token, @options[:qualys_page_size].to_i, page)`

It's true there shouldn't be a situation with nil. But I was thinking more like a safe guard so we don't ingest it? In case the findings_response give us back some nil data? 

Or if it'll make more sense to update the code here 
from: 

`print_debug "there was a problem with the RESPONSE" if findings_response.nil?`

to 

`return if findings_reponses.nil? `

??

tasks/connectors/qualys_was/qualys_was.rb
effany: discussion above :D

tasks/connectors/qualys_was/qualys_was.rb
dianaroz: I think the UI explanation is not needed since the data would be coming from the same place (SVD description/solution) and this is only impacting SVD selection process.

tasks/connectors/qualys_was/qualys_was.rb
effany: remove the safeguard nil line as discussed yesterday :)

sjbridge (approved):
LGTM

tasks/connectors/qualys_was/README.md
```diff
@@ -46,9 +46,10 @@ Run the QualysWas task following the guidelines on the main [toolkit help page](
 | kenna_api_host | hostname | false | Kenna API Hostname |
 | kenna_connector_id | integer | false | If set, we'll try to upload to this connector |
 | output_directory | filename | false | Will alter default filename for output. Path is relative to #{$basedir} |
+| match_ScannerIdentifier_with_vuln_def_name | boolean | false | If true, the finding scanner_identifier and the vuln_def name will match the format 'QID - ID - Name'. If a connector first ran without this parameter and it is enabled in a later run, it will result in automatic resolution of the existing findings with the previous format ('QID - ID') and creation of new findings with the new format within CVM. |
```
dianaroz, line 49: remember to update readme with the new parameter name

tasks/connectors/qualys_was/qualys_was.rb
```diff
@@ -230,7 +235,12 @@ def domain_detail(find_from)
       end
 
       def name(find_from)
-        "#{find_from['qid']}-#{find_from['name']}"
+        if @options[:match_finding_with_vuln_def]
+          structured_name = [find_from['qid'], find_from['id'], find_from['name']].compact.join('-')
+          structured_name unless structured_name.empty?
```
dianaroz, line 240: In the odd case that structured_name was empty, this would result in returning nil for finding scanner_id and vuln_def name. I still think that if we are checking for nil/empty values in toolkit that we need to make sure we are not ingesting nil values into Kenna.

dianaroz (dismissed):

tasks/connectors/qualys_was/README.md
effany: opps thanks! Will update that

tasks/connectors/qualys_was/qualys_was.rb
effany: I agree with that. However, doing the  nil value check in a separate pull request could make more sense? Since this pull request is addressing the issue on mismatch between svd and findings.

effany: Added the safeguard check in the code

tasks/connectors/qualys_was/qualys_was.rb
```diff
@@ -123,6 +128,7 @@ def run(opts)
                   max_records = 0
                 end
                 find_from = data["Finding"]
+                abort("Error: No finding data found. Exiting the process") if find_from.nil?
```
sjbridge, line 131: We wanted to avoid nil vuln_def name because importing that would cause an entire connector run to fail. By `abort`-ing here we are still failing an entire toolkit task run if there is one bad finding. We want to avoid that. I would _only_ do `print_error()` when we are missing this data.

Also if we want to stop execution, there's a `fail_task` helper method we can use: https://github.com/KennaSecurity/toolkit/blob/main/lib/helpers.rb#L77

sjbridge (changes requested):

tasks/connectors/qualys_was/qualys_was.rb
effany: Got it! I will refactor the code :) Thank you

tasks/connectors/qualys_was/qualys_was.rb
```diff
@@ -104,7 +109,7 @@ def run(opts)
         while more_records == true
           findings_response = qualys_was_get_webapp_findings(token, @options[:qualys_page_size].to_i, page)
           more_records = findings_response["ServiceResponse"]["hasMoreRecords"] == "true"
-          print_debug "there was a problem with the RESPONSE" if findings_response.nil?
+          fail_task "there was a problem with the RESPONSE. Nil value provided" if findings_response.nil?
```
sjbridge, line 112: I think it's fine to fail if we're missing the entire page, but I would move it up one line. 111 will fail with an `NoMethodError: undefined method `[]' for nil:NilClass` error

sjbridge (changes requested):

tasks/connectors/qualys_was/qualys_was.rb
effany: got it! thanks!

sjbridge (approved):

PR #533, opened 2024-02-22
Custom properties support for armis connector

parthmodi2-crest:
Adds custom properties support for armis connector

- `customProperties` field from armis device gets mapped as asset tags.


jgarber-cisco (approved):
LGTM!

PR #539, opened 2024-03-22
SUP-1541 Fix for undeclared variable on AWS_guardduty.rb toolkit task

jesutorrr:
### **Problem**
Output_path variable wasn't declared and it is generating errors 

### **Solution**
Using "#{output_dir}/#{filename}" instead

[SUP-1541]: https://kennasecurity.atlassian.net/browse/SUP-1541?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/tasks/connectors/aws_guardduty/aws_guardduty.rb
+++ b/tasks/connectors/aws_guardduty/aws_guardduty.rb
@@ -163,7 +163,7 @@ module Kenna
         ####
         ### Finish by uploading if we're all configured
         ####
-        return unless kenna_connector_id && kenna_api_host && kenna_api_key
+        return unless @options[:kenna_connector_id] && @options[:kenna_api_host] && @options[:kenna_api_key]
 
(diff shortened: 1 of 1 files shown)


effany (commented):
This looks like it'll work to me :slightly_smiling_face:

Just wondering, why we're not using the "kdi_upload" method in the kdi_helpers here ( which include the print_good statement itself) instead create our own output dir in this toolkit task and do exactly what the kdi_upload function is doing?

Wondering if this code can be refactor to be more slim somehow?  :smile: ( but I may be overlooking something )

https://github.com/KennaSecurity/toolkit/blob/22e184497a64ffff07e076accb6d79fd2bec46ae/lib/kdi/kdi_helpers.rb#L205

Thanks!

dianaroz (approved):
looks good

PR #541, opened 2024-04-16
SUP-1567 Undefined method ip_v4_addresses for NilClass

dianaroz:
## SUP-1567

# Problem
AWS lambda findings are not supported by the task but cause it to fail for all findings

# Solution
Filter findings by supported resources (container and ec2)


sjbridge (dismissed):
Looks like the limitation is already documented, so this LGTM

jgarber-cisco (dismissed):

jgarber-cisco (approved):
Nice regeneration for tags

PR #542, opened 2024-04-19
SUP-1569-toolkit error out if findings are empty 

effany:
Currently we don't have the check in place if there's empty qid for a single finding, this will result in error in the toolkit task. 

We should skip creating the finding record if the record is invalid for smoother user experiences.


slubeckatkenna (approved):
Client testing was successful, no errors returned.

sjbridge (commented):
We should also be printing some output for the user if a response page is missing data. Otherwise LGTM

PR #544, opened 2024-04-30
Asimily connector for Cisco VM toolkit

jgarber-cisco:


jgarber-cisco:
@pranav-asimily The lint check is failing. If you can fix up with `bundle exec rubocop -A`.

dianaroz (approved):
Looks good, I'd just add an explanation/example of how the filters should be submitted in the readme file, it looks like that could cause some issues if the right format is not used (i.e. if there are unexpected characters in unexpected places in the string) Or we might want to validate the filters before proceeding with the API call?

jgarber-cisco:
Great point, @dianaroz. I'll merge this and then fix that up. Thank you so much for reviewing!

Issue #550, opened 2024-06-19
Add support to Aqua SaaS


jerbia:
The Aqua connector code only supports connecting to an on-prem Aqua.

For Aqua SaaS deployment there is a need to use a different authentication method (different API).

Is it possible to work together on making these changes in the Aqua connector?

PR #552, opened 2024-07-09
Add regions to aqua

Ukhanskyi:
[Issue](https://github.com/KennaSecurity/toolkit/issues/550)

Add  SaaS  regions to Aqua connector

### DIFF
--- a/tasks/connectors/aqua/lib/aqua_helper.rb
+++ b/tasks/connectors/aqua/lib/aqua_helper.rb
@@ -81,14 +81,7 @@ module Kenna
 
         return unless response
 
-        wp_url = "https://#{JSON.parse(response.body).dig('data', 'ese_url')}"
-
-        if wp_url
-          print_debug("Workload Protection URL retrieved successfully")
-        else
-          print_error("Failed to retrieve Workload Protection URL")
-        end
-        wp_url
+        "https://#{JSON.parse(response.body).dig('data', 'ese_url')}"
       end
 
       def safe_http_get(url, headers)
--- a/tasks/connectors/aqua/lib/aqua_helper.rb
+++ b/tasks/connectors/aqua/lib/aqua_helper.rb
@@ -28,14 +28,13 @@ module Kenna
         @cloud
       end
 
-      def select_region_urls
+      def region_urls
         region = REGION_URLS.keys.find { |key| @aqua_url =~ Regexp.new(key) } || "default"
         REGION_URLS[region]
       end
 
       def aqua_get_token
         if cloud?
-          region_urls = select_region_urls
           get_token(region_urls[:auth], @username, @password)
         else
           get_token("#{@aqua_url}/api/v1/login", @username, @password)
@@ -76,7 +75,6 @@ module Kenna
 
       def get_wp_url(token)
         print_debug "Getting Workload Protection URL"
-        region_urls = select_region_urls
         headers = { "Authorization" => "Bearer #{token}",
                     "Content-Type" => "application/json" }
         response = safe_http_get(region_urls[:wp], headers)
--- a/tasks/connectors/aqua/lib/aqua_helper.rb
+++ b/tasks/connectors/aqua/lib/aqua_helper.rb
@@ -1,53 +1,65 @@
 # frozen_string_literal: true
 
 require "json"
+require "uri"
 
 module Kenna
   module Toolkit
     module AquaHelper
-      def select_region_urls(aqua_url)
-        case aqua_url
-        when /eu-1\.cloud\.aquasec\.com/
-          @saas_auth_url = "https://eu-1.api.cloudsploit.com/v2/signin"
-          @wp_url_api = "https://prov-eu-1.cloud.aquasec.com/v1/envs"
-        when /asia-1\.cloud\.aquasec\.com/
-          @saas_auth_url = "https://asia-1.api.cloudsploit.com/v2/signin"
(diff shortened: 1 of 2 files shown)


Ukhanskyi:
@jgarber-cisco, could I ask you review code one more time? Seems that this is final changes.

tasks/connectors/aqua/lib/aqua_helper.rb
```diff
@@ -5,10 +5,26 @@
...
+        when /asia-1\.cloud\.aquasec\.com/
+          @saas_auth_url = "https://asia-1.api.cloudsploit.com/v2/signin"
+          @wp_url_api = "https://prov-asia-1.cloud.aquasec.com/v1/envs"
+        when /ap-2\.cloud\.aquasec\.com/
+          @saas_auth_url = "https://ap-2.api.cloudsploit.com/v2/signin"
+          @wp_url_api = "https://prov-ap-2.cloud.aquasec.com/v1/envs"
+        else
+          @saas_auth_url = "https://api.cloudsploit.com/v2/signin"
+          @wp_url_api = "https://prov.cloud.aquasec.com/v1/envs"
+        end
```
jgarber-cisco, line 22: Instead of a case statement, these should be stored in a frozen hash, like:
```ruby
REGION_URLS = {
  "eu-1" => {auth: "https://eu-1.api.cloudsploit.com/v2/signin", wp: "https://prov-eu-1.cloud.aquasec.com/v1/envs"},
  "asia-1" => {auth: "https://asia-1.api.cloudsploit.com/v2/signin", wp: "https://prov-asia-1.cloud.aquasec.com/v1/envs"},
  ...
}.freeze
``` 
Then you can write accessor methods that look up auth and wp URLs based on a saved region (the key).

tasks/connectors/aqua/lib/aqua_helper.rb
```diff
@@ -21,14 +37,14 @@ def get_token_from_on_prem(aqua_url, username, password)
...
-        get_token(SAAS_AUTH_URL, username, password)
+        get_token(@saas_auth_url, username, password)
       end
 
       def get_token(auth_url, username, password)
         print_debug "Getting Auth Token from #{auth_url}"
 
         headers = { "Content-Type" => "application/json" }
-        payload = if auth_url == SAAS_AUTH_URL
+        payload = if auth_url == @saas_auth_url
```
jgarber-cisco, line 47: Instead of comparing strings to figure out if this is cloud-based or on-prem, I'd suggest doing a check on initialization and setting an instance variable that has a reader method. Maybe `cloud?` or `saas?`. That way a typo in one of the strings in your case statement won't prevent this check from working. It could be a headache to debug why saas won't auth with an id param but only in one region.

tasks/connectors/aqua/lib/aqua_helper.rb
```diff
@@ -5,10 +5,26 @@
...
+          @saas_auth_url = "https://api.cloudsploit.com/v2/signin"
+          @wp_url_api = "https://prov.cloud.aquasec.com/v1/envs"
+        end
+      end
 
       def aqua_get_token(aqua_url, username, password)
+        select_region_urls(aqua_url)
+
         if cloud_url?(aqua_url)
           get_token_from_cloud(username, password)
```
jgarber-cisco, line 29: You only need to `select_region_urls` if it's a cloud connector, right? That could move inside this if.

tasks/connectors/aqua/lib/aqua_helper.rb
```diff
@@ -59,7 +75,7 @@ def get_token(auth_url, username, password)
       def get_wp_url(token)
         print_debug "Getting Workload Protection URL"
         headers  = { "Authorization" => "Bearer #{token}", "Content-Type" => "application/json" }
-        response = safe_http_get(WP_URL_API, headers)
+        response = safe_http_get(@wp_url_api, headers)
```
jgarber-cisco, line 78: Use the wp url accessor I suggested

tasks/connectors/aqua/lib/aqua_helper.rb
jgarber-cisco: Or every call to a url accessor checks if it's cloud. I realize some of my suggestions are mutually exclusive. Just trying to draw your attention to some readability/maintainability issues.

jgarber-cisco (changes requested):
At least clean up that case statement. YMMV on the other suggestions

tasks/connectors/aqua/lib/aqua_helper.rb
jgarber-cisco: Or if they all follow a pattern, you can build the URL using the `URI` class you're already using to parse the host. Then you don't have to repeat the parts of the string that are in common.

Ukhanskyi:
@jgarber-cisco, seems to me that I have corrected your comments. Could you check if I'm missing something?

jgarber-cisco:
@Ukhanskyi Nice work. I tightened it up a bit more, but I can't test it and there are no specs for this task. Mind running through it? If it works, then I'm good with this PR.

jgarber-cisco (approved):

Ukhanskyi:
> @Ukhanskyi Nice work. I tightened it up a bit more, but I can't test it and there are no specs for this task. Mind running through it? If it works, then I'm good with this PR.

I am very grateful to you for your help and review! In my opinion, everything works fine
image

PR #558, opened 2024-10-07
SUP-1725 Fixing Toolkit Specs

jesutorrr:
SUP-1725

The Snyk toolkit task is creating issues with the specs.
Also fixing CVE-2024-43398 vulnerability


spec/tasks/connectors/snyk_v2/snyk_v2_stubs.rb
```diff
@@ -15,7 +15,7 @@ def stub_projects_request
   end
 
   def stub_issues_request
-    stub_request(:get, Addressable::Template.new("https://api.eu.snyk.io/rest/orgs/{orgId}/issues?created_after=2024-07-14T00:00:00Z&created_before=2024-08-13T00:00:00Z&limit=10&version=2024-04-29"))
+    stub_request(:get, Addressable::Template.new("https://api.eu.snyk.io/rest/orgs/{orgId}/issues?created_after=2024-09-07T00:00:00Z&created_before=2024-10-07T00:00:00Z&limit=10&version=2024-04-29"))
```
dianaroz, line 18: If I understand correctly, the specs are failing because we calculate the from_date of the querystring based on "today". This means that even if this is modified, the specs will continue to fail if they run any other day. 
I would suggest to use Timecop.freeze like we are doing in github_dependabot task:
https://github.com/KennaSecurity/toolkit/blob/980102c61ecd1bfb4706df64ef499b252ea63b51/spec/tasks/connectors/github_dependabot/github_dependabot_spec.rb#L37-L43

It might require some changes in the actual task as well:
https://github.com/KennaSecurity/toolkit/blob/980102c61ecd1bfb4706df64ef499b252ea63b51/tasks/connectors/snyk_v2/snyk_v2_task.rb#L196-L198
```      
@retrieve_from = @options[:retrieve_from]
@from_date = (Time.now - @retrieve_from.to_i.days).utc
@to_date = Time.now.utc
```

tasks/connectors/snyk_v2/snyk_v2_task.rb
```diff
@@ -23,12 +23,12 @@ def self.metadata
               type: "date",
               required: false,
               default: 30,
-              description: "default will be 30 days before today format: YYYY-MM-DD" },
+              description: "Default will be 30 days before today format: YYYY-MM-DD" },
```
dianaroz, line 26: This is not used as a date in the code, it should be defined as integer and the description should make it clear that it is meant to be a number of days.
https://github.com/KennaSecurity/toolkit/blob/980102c61ecd1bfb4706df64ef499b252ea63b51/tasks/connectors/snyk_v2/snyk_v2_task.rb#L197

dianaroz (changes requested):

spec/tasks/connectors/snyk_v2/snyk_v2_stubs.rb
jesutorrr: Thanks for the comment,
I updated the code following your advice!

ababhis2 (approved):
LGTM

PR #566, opened 2025-01-01
Synack task for Kenna toolkit

synack-akozynets:


gcopeland2023 (approved):
passed automated checks. script is in use at customer.

gcopeland2023:
@cmansfield85 sorry for the intrusion. Synack @synack-akozynets has submitted a PR for a new connector in the Kenna toolkit. This is our first submission to the toolkit. We assumed someone from Cisco would provide an approving review. But nothing is happening. We did try reaching out to our Kenna partners contacts but they are not responding. Since i see you active recently in the KennaSecurity/toolkit i was hoping you could review/approve, or advise whom to ask. Thank you in advance for any help.

jgarber-cisco:
Hello, @synack-akozynets and @gcopeland2023! Sorry this has taken so long to be reviewed. We'd like the task to have specs before we merge it in, but that's something we'll gladly take on since we [just started requiring it](https://github.com/KennaSecurity/toolkit/pull/571). Do you have a demo or test Synack account we could use to write API mocks for the specs? You can reach me at jagarber at cisco.com.

gcopeland2023:
@jgarber-cisco thank you for the update. Yes, we can get you access to to a test instance to query our API. I will follow-up via email.

ababhis2 (dismissed):

tasks/connectors/synack/synack.rb
```diff
@@ -41,7 +41,7 @@ def self.metadata
             { name: "asset_defined_in_tag",
               type: "boolean",
               required: false,
-              default: true,
+              default: false,
```
synack-akozynets, line 44: this is correct, thank you. The default should be false

tasks/connectors/synack/synack.rb
```diff
@@ -70,7 +70,7 @@ def run(opts)
         client = Kenna::Toolkit::Synack::SynackClient.new(@synack_api_host, @synack_api_token, @asset_defined_in_tag)
 
         print_good "Attempting to fetch vulns from Synack at #{@synack_api_host}"
-        synack_vulnerabilities = client.fetch_synack_vulnerabilities
+        synack_vulnerabilities = client.fetch_synack_vulnerabilities(page_size: @batch_size)
```
synack-akozynets: This is not going to work well. The batch_size parameter is only to limit how many vulns maximum can be added to a file uploaded to Kenna. This has nothing to do with Synack.
And, what is more important, the synack API has a maximum for that parameter and it is 50. That's why it was hardcoded. Allowing this to be a parameter in requests to Synack is wrong, and won't work as expected.

jgarber-cisco: Okay, then maybe I'll drop `batch_size` and leave `kenna_batch_size` since that's what controls the size of file uploaded to Kenna in other connectors.

Just for my info, do you have a link to the API docs for this endpoint? I couldn't find any with a quick search.

jgarber-cisco: Looks like you interpolated it and made it a constant [here](https://github.com/synack/Azure-Sentinel/blob/eeb21ccc052df20afd1f83b16e57172d65a539bb/Solutions/Synack/Integrations/AzureFunctionSynack/synack-service.js#L63), so perhaps it's good as it is with a default at 50. I might change it to `page_size` to align with your API and avoid confusion with `kenna_batch_size`.

tasks/connectors/synack/synack.rb
```diff
@@ -43,11 +43,11 @@ def self.metadata
               required: false,
               default: false,
               description: "If set to true, will only process Synack vulnerabilities tagged with special asset-defining tags." },
-            { name: "batch_size",
+            { name: "page_size",
```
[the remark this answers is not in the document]
jgarber-cisco: Roger that. I removed `page_size` from the user-configurable options. Thanks!

gcopeland2023 (approved):
Thank You

jgarber-cisco (approved):

PR #574, opened 2025-04-04
SP-1778_add_timeout_retry_error

effany:
https://kennasecurity.atlassian.net/jira/software/c/projects/SUP/boards/58?selectedIssue=SUP-1778

Following above ticket, we notice that we didn't implement a timeout retry error, hence when there was a timeout on the task, we abort it. Adding this code in for completion

### DIFF
--- a/lib/http.rb
+++ b/lib/http.rb
@@ -71,6 +71,14 @@ module Kenna
             sleep(15)
             retry
           end
+        rescue TimeoutLLError => e
+          log_exception(e)
+          if retries < max_retries
+            retries += 1
+            print "Retrying!"
+            sleep(15)
+            retry
+          end
         end
 
         def http_post(url, headers, payload, max_retries = 5, verify_ssl = true)
--- a/spec/rspec_helper.rb
+++ b/spec/rspec_helper.rb
@@ -6,7 +6,7 @@ require 'pry-byebug'
 
 require 'simplecov'
 SimpleCov.start do # Must come before application code is loaded
-  add_filter ['/initialize/', '/scripts/', '/spec/', '/util/']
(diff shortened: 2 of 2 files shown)


ababhis2 (approved):
LGTM

lib/http.rb
```diff
@@ -71,6 +71,14 @@ def http_get(url, headers, max_retries = 5, verify_ssl = true)
             sleep(15)
             retry
           end
+        rescue TimeoutLLError => e
+          log_exception(e)
+          if retries < max_retries
+            retries += 1
+            print "Retrying!"
+            sleep(15)
+            retry
```
jgarber-cisco, line 80: Could eliminate a lot of redundant code by supplying multiple exceptions to `rescue` since the bodies are the same.

jgarber-cisco: At a glance, 6 of these are nearly identical! You could eliminate 7 * 5 = 35 lines of code.

> You can rescue a specific set of exception classes (and their subclasses) by listing them after rescue:

```
begin
  # ...
rescue ArgumentError, NameError
  # handle ArgumentError or NameError
end
```
https://docs.ruby-lang.org/en/3.0/syntax/exceptions_rdoc.html

jgarber-cisco (changes requested):

lib/http.rb
effany: Hello, thanks for the review!
We will be refactoring the code in a separate ticket here
https://kennasecurity.atlassian.net/jira/software/c/projects/SUP/boards/58?selectedIssue=SUP-1793

jgarber-cisco (approved):
The rescues in http_get will be DRYed up later. 👍

PR #577, opened 2025-04-16
Test GH Actions

jgarber-cisco:
Seeing if they run on every commit to a PR


jgarber-cisco:
Only two checks were run on `6fce49a`. Why?!

jgarber-cisco:
Seems to be going okay since I deleted the CodeQL action that duplicated the built-in CodeQL.

PR #578, opened 2025-04-16
Update super-linter

jgarber-cisco:
Using the latest version and config from their readme: https://github.com/super-linter/super-linter?tab=readme-ov-file


mattahill (approved):
quick glance looks ok to me

PR #584, opened 2025-05-07
SUP-1778-add the econnreset error retry

effany:
Add econnreset retry logic in.


lib/http.rb
ababhis2, line 1: @jgarber-cisco  any reason to not update nokogiri, I am like lets give it a shot

jgarber-cisco (approved):
Hope you figure out the exception that helps this customer.

PR #590, opened 2025-06-30
SUP-1821-scordcard-task-404-error

effany:
customer's task is facing 404 error due to endpoint change. 

However, I also notice that how we're retrieving data is fundamentally wrong anyway....


jgarber-cisco:
Would it be too much trouble to add a spec in https://github.com/KennaSecurity/toolkit/tree/main/spec/tasks/connectors/digital_footprint? I almost said, "we need to skip that codecov check again," but then I thought maybe the pain it causes is a good thing! Once there's the stub of a spec there, it becomes easier to add coverage for scorecard every time we touch it.

ababhis2 (dismissed):

PR #591, opened 2025-06-30
- improved readme file as discussed on a call with Kenna/Cisco team

synack-akozynets:


jgarber-cisco (approved):
Thanks for updating the documentation with support information and a refined description!

PR #596, opened 2025-07-10
Sup-1821-toolkit _scorecard_fix_v3

effany:
customer's task is facing 404 error due to endpoint change.

However, I also notice that how we're retrieving data is fundamentally wrong anyway....


tasks/connectors/digital_footprint/security_scorecard/lib/client.rb
```diff
@@ -38,13 +38,18 @@ def companies_by_portfolio(portfolio_id)
...
+            JSON.parse(response.body)
+          rescue JSON::ParserError
+            { "error" => "Invalid JSON response", "raw_body" => response.body }
+          end
         end
 
         def issues_by_type_for_company(company_id, itype = "patching_cadence_low")
           endpoint = "#{@baseapi}/companies/#{company_id}/issues/#{itype}"
           response = http_get(endpoint, @headers, 0)
-          JSON.parse(response.body.to_s) unless response.nil?
```
abhiuee, line 47: do we need to remove `to_s` from here?

jgarber-cisco (dismissed):
I'd prefer the speculative nil checks and parser error handling was removed, but it's in good company with what already exists.

ababhis2 (dismissed):

PR #614, opened 2026-06-12
SUP-1886-update veracode retry on 504

effany:
This PR updates Veracode retry handling for 504 Gateway Timeout responses.

When a request to Veracode returns HTTP 504, the connector now retries the request up to 2 times, with a fixed 30-second interval between attempts.

This change improves resiliency against temporary Veracode-side timeouts while keeping retry behavior controlled and predictable.

https://kennasecurity.atlassian.net/jira/software/c/projects/SUP/boards/58?selectedIssue=SUP-1886

### DIFF
--- a/tasks/connectors/veracode/lib/veracode_client.rb
+++ b/tasks/connectors/veracode/lib/veracode_client.rb
@@ -17,7 +17,6 @@ module Kenna
           max_interval: 30,
           backoff_factor: 1,
           methods: %i[get],
-          exceptions: [],
           retry_statuses: [504]
         }.freeze
         MAX_RETRIES_504_ONLY = 2
--- a/tasks/connectors/veracode/lib/veracode_client.rb
+++ b/tasks/connectors/veracode/lib/veracode_client.rb
@@ -12,6 +12,15 @@ module Kenna
         CWE_PATH = "/appsec/v1/cwes"
         FINDING_PATH = "/appsec/v2/applications"
         REQUEST_VERSION = "vcode_request_version_1"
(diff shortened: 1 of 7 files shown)


tasks/connectors/veracode/lib/veracode_client.rb
```diff
@@ -82,7 +90,7 @@ def process_paged_findings(app_guid, scan_type, &)
         def get_paged_results(url)
           next_page = url
           until next_page.nil?
-            response = http_get(next_page, {}, hmac_client: self)
+            response = http_get(next_page, {}, MAX_RETRIES_504_ONLY, true, hmac_client: self, retry_options: RETRY_OPTIONS_504_ONLY)
```
ababhis2: can you please elaborate what this is going to do?

504 was already part of the http.rb retry options. how does this all work

The way I am reading this is now instead of retrying 5 times we will be retrying 2 times. is that true?

ababhis2 (commented):
Need some clarity on the work

tasks/connectors/veracode/lib/veracode_client.rb
effany: yes, passing MAX_RETRIES_504_ONLY = 2 means this call now allows 2 retries, not 5. That is 2 retries after the initial request, so 3 total attempts. Also, specifically for Veracode , it uses the longer 30-second interval settings from RETRY_OPTIONS_504_ONLY

ababhis2 (approved):
nit: more testing but this is looking good now. 

Approved

